Issue #16, opened 2015-11-23
Focused tests/contexts with `fit()` and `fdescribe()`


greghaskins:
Similar to Jasmine's API: http://jasmine.github.io/2.1/focused_specs.html

greghaskins:
Not sure whether we want to settle on Jasmine style `fit()` or more like Mocha with `it.only()`

Should have some consistency with ignoring tests (#5, `xit()` vs `it.skip()` or something)

kdroberson:
+1 !

jonstorer:
+1! Any timeline on this? Or direction for a PR?

greghaskins:
I've started on a first-pass implementation for `xit` / `xdescribe` that is related to the focused tests feature as well. I could ship one or the other in the next week or so. 

@jonstorer @kdroberson would focused tests or skipped tests be more useful in the near term? 

Moving forward, there are some structural changes that need to happen to make it easier for people other than me to contribute PRs. Some of the implementation is a little too clever for its own good, and the tests don't have a very consistent style/ format.

jonstorer:
@greghaskins tbh, focusing would be better up front.

greghaskins:
Agreed. Especially since there's no easy way to run just a single test (or subset of tests). I'll implement `fit()` and `fdescribe` first. Perhaps even just `fit()` to start with, in case the context-level logic gets tricky.

In terms of API, I've been essentially copying Jasmine as much as possible, mostly because it is simpler than Mocha and RSpec which each have numerous ways to accomplish the same thing. I am open to feedback/ideas, though.

jonstorer:
Get a pull request started and I'll happily provide feedback. I agree with the Jasmine approach since simpler is always better.

I'm new to a team of Java developers, but I'm coming from the Ruby/JavaScript world. Between us we should be able to provide help. 

Thanks for keep us in the loop.

greghaskins:
I've got PR #23 out there now for `fit` and `fdescribe`. If you're interested, I'd welcome any feedback you have. Otherwise I'll merge and push a `0.6.0` release by midweek.

greghaskins:
Merged #23 and released [0.6.0](https://github.com/greghaskins/spectrum/releases/tag/0.6.0) with `fdescribe` and `fit`. It's available now on jcenter.

PR #23, opened 2016-02-01
Feature: focused specs with `fit` and `fdescribe`

greghaskins:
Essentially, just adding a `focus()` method to `Suite` and `Spec`. The tricky bit was propagating that all the way up the hierarchy in a non-ugly way. It would be nice not to have doubly-linked parent-child and  child-parent pointers, but I couldn't find a clean way to avoid it.

To minimize the inappropriate intimacy, in this implementation, I created explicit `Parent` and `Child` interfaces, so that child and leaf nodes have only minimal knowledge of what their parent can do (currently: just `focus(Child)`).


greghaskins:
This PR will resolve #16

PR #30, opened 2016-05-22
Gradle caching on Travis CI

greghaskins:
Turning on Travis-CI caching according to the [official instructions](https://docs.travis-ci.com/user/languages/java#Caching). This should make builds on `master` and on PRs run a lot faster.


greghaskins:
After experimenting with this, it doesn't seem to be helping much. Build times are still around 2 minutes. Perhaps the overhead of the S3 transfers, along with the per-branch isolation, wipe out the potential upside (https://docs.travis-ci.com/user/caching/#How-does-the-caching-work%3F).

It's not worth the additional build complexity to keep it in, so I'm closing this PR.

Issue #6, opened 2014-12-03
xdescribe() for ignored/pending suite


greghaskins:
```java
xdescribe("A spec that is ignored", () -> {
    // will not run

    it("might have tests, but they will also not run", () -> {
        // will not run
    });
});

greghaskins:
Closed by #33

PR #33, opened 2016-05-27
Adding `xit` and `xdescribe` to allow Ignored specs.

### LINKED ISSUE #5: xit() for ignored/pending test
```java
xit("An ignored test", () -> {
    // will not run
});
```

tlabeeuw:
Resolves #5 and #6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,42 @@ describe("Focused specs", () -> {
 });
 ```
 
+### Ignored Specs
+
+You can ignore the runner on particular spec with `xit` or a suite with `xdescribe` so that only those specs get executed.
+
+> from [IgnoredSpecs.java](src/test/java/specs/IgnoredSpecs.java)
+
+```java
+describe("Ignored specs", () -> {
+
+    xit("is ignored and will not run", () -> {
+        throw new Exception();
+    });
+
+    it("is not ignored and will run", () -> {
+        assertThat(true, is(true));
+    });
+
+    xdescribe("an ignored suite", () -> {
+
+        it("will not run", () -> {
+            throw new Exception();
+        });
+
+        describe("with nesting", () -> {
+            it("all its specs", () -> {
+                throw new Exception();
+            });
+            
+            fit("including focused specs", () -> {
+                throw new Exception();
+            });
+        });
+    });
+});
+```
+
 ## Supported Features
 
 Spectrum moving toward a `1.0` release with close alignment to Jasmine's test declaration API. The library already supports a nice subset of those features:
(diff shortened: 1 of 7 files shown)


greghaskins:
@tlabeeuw nice work on this PR! It looks really great, including the tests and docs too.

One question on your implementation: why did you decide to have `Spec` objects know about their parent's `isIgnored` state? One alternative might have been for the `Suite` to automatically ignore all its children somehow, perhaps by adding a check [here in `run()`](https://github.com/tlabeeuw/spectrum/blob/0f2c16e6de128c95ea4b17dfc8e0b95b4ae33e6f/src/main/java/com/greghaskins/spectrum/Suite.java#L121) or [here in `runChild()`](https://github.com/tlabeeuw/spectrum/blob/0f2c16e6de128c95ea4b17dfc8e0b95b4ae33e6f/src/main/java/com/greghaskins/spectrum/Suite.java#L135). Then you might not need the [`isIgnored()` method on `Parent`](https://github.com/tlabeeuw/spectrum/blob/0f2c16e6de128c95ea4b17dfc8e0b95b4ae33e6f/src/main/java/com/greghaskins/spectrum/Parent.java#L7) at all.

Mostly I'm just curious if there's a way to make the `Parent`/`Child` relationship even simpler (i.e. fewer methods).

greghaskins:
When this is merged, it will close #5 and #6.

tlabeeuw:
@greghaskins Thanks for merging. 

I felt that putting the logic in [Suite.java#run()](https://github.com/greghaskins/spectrum/blob/master/src/main/java/com/greghaskins/spectrum/Suite.java#L119-L140), violated the "tell don't ask" design principal. This way the parent just tells the child (be it a leaf node or not) to run, and it's up to the child to actually decide if it wants to run. That way less branching in the code.

tlabeeuw:
PS: Slightly off topic but I'd maybe even suggest we could carry this further. Change the `run()` method too:

```java
  @Override
  public void run(final RunNotifier notifier) {
    this.children.forEach((child) -> child.run(notifier));
  }
```

And delete the `runChildren()`, `runChild()`, `testCount()` and `focusedChildren`, `addSpec()`, `createSpec()`, and others. Instead have a Builder create `IgnoredSpec` (for non-focused & ignored), `Spec` and `Suite`. This way the Parent, Child, Suite, IgnoredSpec, and Spec objects are all very simple objects, without a single `if` branching statement.

Sorry don't know if that was actually readable. Maybe should just do a PR to demonstrate.

greghaskins:
I like where you're going with that, feels like it could be a lot simpler. I wonder if that could also lead to a cleaner way for us to handle the focusing scenario where `focus()` essentially would ignore the whole world, except any specs that were also focused, but only if they themselves weren't in an `xdescribe`.

I played around with a couple ideas, but got caught by some of the edge cases you (thankfully!) added tests for (e.g. normal, focused, and ignored specs across completely different suites). I'd be happy to check out a PR if you take a stab it.

greghaskins:
Also, @tlabeeuw, you should link your email to your GitHub account so you show up in the [list of contributors](https://github.com/greghaskins/spectrum/graphs/contributors)!

https://help.github.com/articles/i-don-t-see-myself-in-the-repository-contributors-graph/#account-linking

Issue #53, opened 2016-11-27
Support for categories/tagging of Specs


ashleyfrieze:
At the moment there is an Rspec like syntax for focusing or ignoring individual nodes within the tree. What would be useful is to enable CI builds that build sections of the tests, or for developers to be able to run WIP tests exclusively while stopping them from running in CI builds.

- Make it easy to selectively run features and scenarios:
  - This probably means a bigger mechanism than the ignore/focus that's presently done within Spectrum
  - For non-trivial situations, this also probably means allowing for tagging of nodes in the tree
  - If there's tagging, then there needs to be a mechanism for selecting the tag
  - The selection mechanism needs to be available at development time and also at CI time
  - Duplicating Cucumber's tag feature is not a perfect solution, but the solution should be a cousin of that

ashleyfrieze:
A few ways it can be done, and my thoughts on them:

* Modifying the call to `describe`/`it` - `xdescribe`/`xit` - this would not work for configuration-based exclusion
* A tag or ignore method call before a describe/it/feature/scenario - e.g. `ignore(); describe(...);` this reads like english and is strongly typed
* Using something in the text of the name of the item - e.g. `"My lovely feature @new"` - this is weakly typed and requires lot of string manipulation
* Calling an ignore function or tagging function after the block - e.g. `describe(...).tag("new")` - this reads backwards

ashleyfrieze:
Proposal.

- Keep `xit`,`xdescribe`,`fit`,`fdescribe` working the way they do now
- Add the following commands which preceed suite and spec declarations and affect the single one after them:
   - `ignore()` - means ignore the next item
   - `focus()` - means add a focus to the next item - so only focused items will run
   - `tag(...)` - means add the following tags to the next item - this will work in conjunction with other mechanisms to either focus or ignore the next item
   - `focusTags(...)` - make any item tagged with any of the provided tags go into focus
   - `requireTags(...)` - make any suite level item NOT tagged with any of the provided tags get ignored
   - `excludeTag(...)` - make anything tagged with any of the provided tags get ignored
- The focusTags/requireTags/excludeTags can be called many times over and the last one called is the winner
- There will also be a `SpectrumOptions` annotation within which the  focusTags/requireTags/excludeTags can be provided programmatically - this would allow for a common base class to have some overall tagging, in addition, the  SpectrumOptions can define which system properties are used to read these tags
[...]

PR #55, opened 2016-11-29
Added a form of spec ignoring with the `pending` function.

ashleyfrieze:
Part of #53 this is a more Jasmine/Rspec way of marking specs as ignored


greghaskins:
Looks good. This is just for specs/steps of course (for those finding this PR later). I have an idea about how to make `pending()` work inside a suite/feature as well, but that can wait for a separate PR.

Issue #49, opened 2016-11-18
@Rule and @ClassRule support


rkawala:
JUnit provides @ClassRule and @Rule annotations, which would let me use the new @SpringBootTest annotation with Spectrum. Is there any plan to provide support for @ClassRule and @Rule? Thanks!

rkawala:
I think it might be as simple as changing Spectrum to extend ParentRunner rather than implementing Runner.

greghaskins:
I like the idea of using existing JUnit extension points like `@Rule` and `@ClassRule` if we can. The tricky bit will be figuring out a clean way to do so. Most of the Spring testing stuff relies on custom subclasses of the base JUnit runner, which would be difficult to integrate directly into Spectrum unless we copy their code.

For example, based on my reading of [the Spring docs](https://spring.io/blog/2016/04/15/testing-improvements-in-spring-boot-1-4#spring-boot-1-4-simplifications), using `@SpringBootTest` requires using their custom runner (`@RunWith(SpringRunner.class)`) and thus you couldn't use `@RunWith(Spectrum.class)`.

greghaskins:
I don't think extending `ParentRunner` is a viable option for Spectrum. Its implementation makes some fundamental assumptions that conflict with Spectrum, like tests being declared as methods, suites being simple Java classes etc.

However, it would be fairly straightforward to add an adapter so that we could use Rules with Spectrum, just declared in a different way. Borrowing some examples from the [JUnit wiki](https://github.com/junit-team/junit4/wiki/Rules) it could look something like this:


```java
describe("a possible way of integrating JUnit rules", () -> {

    Supplier<TemporaryFolder> tempFolder = rule(() -> new TemporaryFolder());

    Supplier<ExpectedException> exception = rule(() -> ExpectedException.none());

    it("counts assets", () -> {
        File icon = tempFolder.get().newFile("icon.png");
        File assets = tempFolder.get().newFolder("assets");
        createAssets(assets, 3);

        DigitalAssetManager dam = new DigitalAssetManager(icon, assets);
        assertEquals(3, dam.getAssetCount());
    });

    it("throws IllegalArgumentException if icon is null", () -> {
        exception.get().expect(IllegalArgumentException.class);
        exception.get().expectMessage("Icon is null, not a file, or doesn't exist.");
        new DigitalAssetManager(null, null);
    });

});
```

I still don't think this will make Spring stuff "just work" but it would provide the beginnings of some extension points.

Thoughts?

[...]

Issue #34, opened 2016-07-08
Mockito annotations support?


ishustava:
We have found that using `@Mock` and `@InjectMocks` annotations don't work with spectrum, resulting in the mock causing a `NullPointerException`.

It seems that the runner needs to do all the injections and mocking pretty early in the test execution, so that all instances are injected and/or mocked.

This issue is somewhat related to [this issue](https://github.com/greghaskins/spectrum/issues/15), but perhaps it's helpful to bring it up here separately.

Thanks for the great framework,

@ishustava && @larham

greghaskins:
Thanks for filing an issue @ishustava and @larham, it is somewhat different from #15 and worth tracking. `@Mock` and `@InjectMocks` only work with `MocktioJUnitRunner`, and probably won't be able to be integrated out-of-the-box with Spectrum.

Making Spectrum easier to work with Mocktio, however, is something that we should consider. It wouldn't look exactly the same, but would provide some of the same convenience. I have been thinking that another helper/wrapper could be useful, perhaps to run in a `beforeEach` block to wire up mocks. That would probably be a related, but separate project. 

Just sketching out ideas, what do you think of something like this?

```java
describe("a cash register", () -> {

	Value<CashRegister> cashRegister = value();
	Value<Drawer> drawer = value();
	Value<BarcodeScanner> scanner = value();

	beforeEach(() -> {
		injectMocks(drawer, scanner).into(cashRegister));
	});
	
	it("should open the drawer to accept payment", () -> {
		cashRegister.acceptPayment();
		
		verify(drawer).open();
	});
	
	it("should read from the barcode scanner", () -> {
		when(scanner.read()).thenReturn(123456);
		
		cashRegister.readNextItem();
		
		expect(cashRegister.lastItem()).toBe(123456);
	});

});
```

larham:
in our attempts so far, injection can be handled with beforeEach(() -> MockitoAnnotations.injectMocks(this));

larham:
```@RunWith(Spectrum.class)
public class SampleMockito {
​
  @Mock
  Environment environment;
​
  @InjectMocks
  Thing subject;
​
  beforeEach(() -> MockitoAnnotations.injectMocks(this));
​
  describe("something", ...)
}
```

larham:
is there any magic that could detect the presence of the Mockito annotations and add that "before" injection under the hood?

greghaskins:
Thanks for sharing the example. I understand better now. 

[...]

Issue #15, opened 2015-11-09
Integration with Mockito


petergphillips:
Hi Greg,

I have been having at Spectrum and really like the way the tests get grouped together.  I wondered if there was any support for mockito annotations, or if you knew a way to get them working.

I have managed to get a Mockito test by explicitly constructing the mocks and then adding a manual call to validateMockitoUsage and reset in the afterEach, it just feels quite hard work.

Normally I would add in a Mockito JUnit rule, annotate each mock and then construct the service in a before method.  Having to manually construct the mocks is not too bad, but It would be great if there was some way to trigger the validate and reset at the end of each test.

Thanks,

Peter.

greghaskins:
Interesting concept. I hadn't even considered how using Mockito might feel with Spectrum. I might put some thought into it if I can get some spare cycles. 

Do you have some sample code I could check out using the technique you described?

petergphillips:
I've taken one of your tests using Mockito and converted it into an example - please see https://github.com/petergphillips/spectrum/commit/bb85dd87968f14d896688f33178bfd63e3aa02eb.  Obviously not a good way to test Spectrum, but should give you an idea of how I had to integrate it...

ashleyfrieze:
#49 should fix this. Once you can use rules to interact with objects in the test class's instance a few opportunities open up.

ashleyfrieze:
I've made a hacked version of rules in Spectrum - see the spec file, rather than the implementation:

https://github.com/ashleyfrieze/spectrum/blob/scratch-implementation-rules/src/test/java/specs/MockitoSpec.java

```java
@RunWith(Spectrum.class)
public class MockitoSpec {
	@Mock
	private SomeInterface mockInterface;

	@InjectMocks
	private SomeClass objectUnderTest;

	@Rule
	public MockitoRule rule = MockitoJUnit.rule();

	{
		describe("A suite which needs mockito", () -> {
			beforeEach(() -> {
				given(mockInterface.getInput()).willReturn("Hello world");
			});

			it("can use the mocks", () -> {
				assertThat(objectUnderTest.getResult(), is("Hello world"));
			});

			it("can use the mocks again", () -> {
				assertThat(objectUnderTest.getResult(), is("Hello world"));
			});

			it("gets a fresh mock each time", () -> {
				verify(mockInterface, never()).getInput();
			});

		});
	}
}
```

[...]

PR #56, opened 2016-11-29
JUnit rules - attempt to implement them in Spectrum

ashleyfrieze:
This appears to work for Mockito and TemporaryFolder as evidenced. It _should_ work for Class rules too and needs roadtesting with Spring.

It should ultimately solve #15 #49 and #34

### DIFF
--- a/src/main/java/com/greghaskins/spectrum/Spec.java
+++ b/src/main/java/com/greghaskins/spectrum/Spec.java
@@ -61,4 +61,9 @@ final class Spec implements Child {
   public boolean isAtomic() {
     return true;
   }
+
+  @Override
+  public boolean isEffectivelyIgnored() {
+    return ignored;
+  }
 }
--- a/src/main/java/com/greghaskins/spectrum/Spectrum.java
+++ b/src/main/java/com/greghaskins/spectrum/Spectrum.java
@@ -7,6 +7,7 @@ import static com.greghaskins.spectrum.model.BlockConfiguration.Factory.focus;
 import static com.greghaskins.spectrum.model.BlockConfiguration.Factory.ignore;
 
 import com.greghaskins.spectrum.internal.LetHook;
+import com.greghaskins.spectrum.internal.junit.Rules;
 import com.greghaskins.spectrum.model.ConstructorBlock;
 import com.greghaskins.spectrum.model.HookContext;
 import com.greghaskins.spectrum.model.IdempotentBlock;
@@ -223,6 +224,30 @@ public final class Spectrum extends Runner {
 
 
   /**
+   * Implements a rules context within the current suite using the given rules class as a mix-in.
+   * These rules will cascade down and be applied at the level of specs or atomic specs.
+   * @param rulesClass type of object to create and apply rules to for each spec.
+   * @param <T> type of the object
+   * @return a supplier of the rules object
+   */
+  public static <T> Supplier<T> applyRules(final Class<T> rulesClass) {
+    return Rules.applyRules(rulesClass);
+  }
+
+  /**
+   * Implements a rules context within the current suite using the given rules class as a mix-in.
+   * These rules will only run for each immediate child of the suite. If the rules should
+   * run fresh in a fresh rules object for each atomic spec
+   * ({@link #it(String, com.greghaskins.spectrum.Block)} etc) then use {@link #applyRules(Class)}
+   * @param rulesClass type of object to create and apply rules to for each spec.
+   * @param <T> type of the object
+   * @return a supplier of the rules object
+   */
+  public static <T> Supplier<T> applyRulesHere(final Class<T> rulesClass) {
+    return Rules.applyRulesHere(rulesClass);
+  }
+
+  /**
    * Declare a {@link com.greghaskins.spectrum.Block} to be run before each spec in the suite.
    *
    * <p>
@@ -292,7 +317,7 @@ public final class Spectrum extends Runner {
    * @param consumer to run each spec block
    */
   public static void aroundEach(ThrowingConsumer<com.greghaskins.spectrum.Block> consumer) {
-    addHook(new HookContext(consumer::acceptOrThrow, getDepth(),
+    addHook(new HookContext(Hook.from(consumer), getDepth(),
         HookContext.AppliesTo.ATOMIC_ONLY, HookContext.Precedence.GUARANTEED_CLEAN_UP_LOCAL));
   }
 
@@ -304,7 +329,7 @@ public final class Spectrum extends Runner {
    * @param consumer to run each spec block
    */
   public static void aroundAll(ThrowingConsumer<com.greghaskins.spectrum.Block> consumer) {
-    addHook(new HookContext(consumer::acceptOrThrow, getDepth(),
+    addHook(new HookContext(Hook.from(consumer), getDepth(),
         HookContext.AppliesTo.ONCE, HookContext.Precedence.OUTER));
   }
 
@@ -329,6 +354,22 @@ public final class Spectrum extends Runner {
     addHook(hookContext);
 
     return letHook;
+  }
+
+  /**
+   * Insert a hook into the current level of definition.
+   * @param hook to insert
+   * @param appliesTo the {@link com.greghaskins.spectrum.model.HookContext.AppliesTo} indicating
+   *                  where the hook is run
+   * @param precedence the importance of the hook compared to others
+   */
+  public static void addHook(final Hook hook, final HookContext.AppliesTo appliesTo,
+      final HookContext.Precedence precedence) {
+    addHook(new HookContext(hook, getDepth(), appliesTo, precedence));
+  }
+
+  private static void addHook(HookContext hook) {
+    getCurrentSuiteBeingDeclared().addHook(hook);
   }
 
   /**
@@ -376,8 +417,10 @@ public final class Spectrum extends Runner {
   private static void beginDefinition(final Suite suite,
       final com.greghaskins.spectrum.Block definitionBlock) {
     getSuiteStack().push(suite);
+
     try {
       definitionBlock.run();
+      addRootJUnitHook(definitionBlock);
     } catch (final Throwable error) {
       suite.removeAllChildren();
       it("encountered an error", () -> {
@@ -399,7 +442,16 @@ public final class Spectrum extends Runner {
     return getSuiteStack().peek();
   }
 
-  private static void addHook(HookContext hook) {
-    getCurrentSuiteBeingDeclared().addHook(hook);
+  /**
+   * If the definition is at the right level and the block is the right sort
+   * of block, then this is might be an opportunity to hook in JUnit rules.
+   * @param definitionBlock the block used to make this suite
+   */
+  private static void addRootJUnitHook(com.greghaskins.spectrum.Block definitionBlock) {
+    if (getDepth() == 1) {
+      if (definitionBlock instanceof ConstructorBlock) {
+        Rules.applyRules(((ConstructorBlock) definitionBlock).get());
+      }
+    }
   }
 }
--- a/src/main/java/com/greghaskins/spectrum/Spectrum.java
+++ b/src/main/java/com/greghaskins/spectrum/Spectrum.java
@@ -224,27 +224,14 @@ public final class Spectrum extends Runner {
 
 
   /**
-   * Implements a rules context within the current suite using the given rules class as a mix-in.
+   * Uses the given class as a mix-in for JUnit rules to be applied.
    * These rules will cascade down and be applied at the level of specs or atomic specs.
    * @param rulesClass type of object to create and apply rules to for each spec.
    * @param <T> type of the object
    * @return a supplier of the rules object
    */
-  public static <T> Supplier<T> applyRules(final Class<T> rulesClass) {
+  public static <T> Supplier<T> junitMixin(final Class<T> rulesClass) {
     return Rules.applyRules(rulesClass);
-  }
-
-  /**
-   * Implements a rules context within the current suite using the given rules class as a mix-in.
-   * These rules will only run for each immediate child of the suite. If the rules should
-   * run fresh in a fresh rules object for each atomic spec
-   * ({@link #it(String, com.greghaskins.spectrum.Block)} etc) then use {@link #applyRules(Class)}
-   * @param rulesClass type of object to create and apply rules to for each spec.
-   * @param <T> type of the object
-   * @return a supplier of the rules object
-   */
-  public static <T> Supplier<T> applyRulesHere(final Class<T> rulesClass) {
-    return Rules.applyRulesHere(rulesClass);
   }
 
   /**
@@ -396,7 +383,7 @@ public final class Spectrum extends Runner {
    * @see org.junit.runner.Runner
    */
   public Spectrum(final Class<?> testClass) {
-    this(Description.createSuiteDescription(testClass), new ConstructorBlock(testClass));
+    this(Description.createSuiteDescription(testClass), createTestClassDefinitionBlock(testClass));
   }
 
   Spectrum(Description description, com.greghaskins.spectrum.Block definitionBlock) {
@@ -420,7 +407,6 @@ public final class Spectrum extends Runner {
 
     try {
       definitionBlock.run();
-      addRootJUnitHook(definitionBlock);
     } catch (final Throwable error) {
       suite.removeAllChildren();
       it("encountered an error", () -> {
@@ -428,6 +414,23 @@ public final class Spectrum extends Runner {
       });
     }
     getSuiteStack().pop();
+  }
+
+  /**
+   * Links the test class construction to JUnit rules implementation.
+   * @param testClass type of the test object
+   * @return a block which when executed will perform test definition against Spectrum
+   *         and also hooks JUnit rule implementation to the definition based on any
+   *         "@Rule" annotations on the members - see {@link Rules}
+   */
+  private static <T> com.greghaskins.spectrum.Block createTestClassDefinitionBlock(
+      final Class<T> testClass) {
+    ConstructorBlock<T> constructTestClass = new ConstructorBlock<>(testClass);
+
+    return () -> {
+      constructTestClass.run();
+      Rules.applyRules(constructTestClass.get());
+    };
   }
 
   private static Deque<Suite> getSuiteStack() {
@@ -440,18 +443,5 @@ public final class Spectrum extends Runner {
 
   private static Suite getCurrentSuiteBeingDeclared() {
     return getSuiteStack().peek();
-  }
-
-  /**
-   * If the definition is at the right level and the block is the right sort
-   * of block, then this is might be an opportunity to hook in JUnit rules.
-   * @param definitionBlock the block used to make this suite
-   */
-  private static void addRootJUnitHook(com.greghaskins.spectrum.Block definitionBlock) {
-    if (getDepth() == 1) {
-      if (definitionBlock instanceof ConstructorBlock) {
-        Rules.applyRules(((ConstructorBlock) definitionBlock).get());
-      }
-    }
   }
 }
--- a/src/main/java/com/greghaskins/spectrum/Suite.java
+++ b/src/main/java/com/greghaskins/spectrum/Suite.java
@@ -200,8 +200,12 @@ class Suite implements Parent, Child {
   }
 
   private void runSuite(final RunNotifier notifier) {
-    hooks.once().sorted()
-        .runAround(description, notifier, () -> runChildren(notifier));
+    if (isEffectivelyIgnored()) {
+      runChildren(notifier);
+    } else {
+      hooks.once().sorted()
+          .runAround(description, notifier, () -> runChildren(notifier));
+    }
   }
 
   private void runChildren(final RunNotifier notifier) {
@@ -209,12 +213,19 @@ class Suite implements Parent, Child {
   }
 
   protected void runChild(final Child child, final RunNotifier notifier) {
-    if (this.focusedChildren.isEmpty() || this.focusedChildren.contains(child)) {
+    if (child.isEffectivelyIgnored()) {
+      // running the child will make it act ignored
+      child.run(notifier);
+    } else if (childIsNotInFocus(child)) {
+      notifier.fireTestIgnored(child.getDescription());
+    } else {
       hooks.forThisLevel().sorted().runAround(child.getDescription(), notifier,
           () -> runChildWithHooks(child, notifier));
-    } else {
-      notifier.fireTestIgnored(child.getDescription());
     }
+  }
+
+  private boolean childIsNotInFocus(Child child) {
+    return !this.focusedChildren.isEmpty() && !this.focusedChildren.contains(child);
   }
 
   private void runChildWithHooks(final Child child, final RunNotifier notifier) {
@@ -245,5 +256,17 @@ class Suite implements Parent, Child {
 
   private String sanitise(final String name) {
     return nameSanitiser.sanitise(name);
+  }
+
+  @Override
+  public boolean isEffectivelyIgnored() {
+    return ignored || !hasANonIgnoredChild();
+  }
+
+  private boolean hasANonIgnoredChild() {
+    return children.stream()
+        .filter(child -> !child.isEffectivelyIgnored())
+        .findFirst()
+        .isPresent();
   }
 }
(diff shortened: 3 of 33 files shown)


ashleyfrieze:
I'll tackle the code coverage after initial feedback on how this goes. I think this solution is easily extended to a more JUnit native look too - for situations where you don't need a fresh instance of the test class, i think you could just do

`applyRules(this)`

or maybe have Spectrum do this for you as the first step, since if "this" doesn't have rules, nothing will be applied.

ashleyfrieze:
So I got the coverage back to 100%... this PR needs more thought and more testing. Note: the mixin classes must be public or JUnit won't play with them.

Note: I stole a lot of the implementation from JUnit BlockJUnitRunner. Knowing where to put it is the hard bit :)

src/main/java/com/greghaskins/spectrum/CompositeTest.java
```diff
@@ -0,0 +1,34 @@
+package com.greghaskins.spectrum;
+
+import org.junit.runner.Description;
+import org.junit.runner.notification.RunNotifier;
+
+/**
+ * Subclass of {@link Suite} that represent the fact that some tests are composed
+ * of interrelated steps which add up to a single test.
+ */
+class CompositeTest extends Suite implements Atomic {
```
ashleyfrieze: Extracted to class as `Suite` was getting big!

src/main/java/com/greghaskins/spectrum/ConstructorBlock.java
```diff
@@ -16,14 +18,19 @@ public void run() throws Throwable {
     try {
       final Constructor<?> constructor = this.klass.getDeclaredConstructor();
       constructor.setAccessible(true);
-      constructor.newInstance();
+      testObject = constructor.newInstance();
```
ashleyfrieze: We'll need this soon... I'm planning a cheat version of the rule stuff where you can use the test class the old way if you like and it just kinda works...

src/main/java/com/greghaskins/spectrum/RuleContextScope.java
```diff
@@ -0,0 +1,46 @@
...
+class RuleContextScope<T> implements Supplier<RuleContext<T>> {
+  private RuleContext<T> ruleContext;
+  private int generation;
+
+  static <T> RuleContextScope<T> ofRecursive(final RuleContext<T> context) {
+    return new RuleContextScope<>(context, 0);
+  }
+
+  static <T> RuleContextScope<T> ofNonRecursive(final RuleContext<T> context) {
+    return new RuleContextScope<>(context, -2);
```
ashleyfrieze: yeah.. -2 - that means it's "my generation" - this needs woyk!

src/main/java/com/greghaskins/spectrum/Spec.java
```diff
@@ -3,7 +3,7 @@
 import org.junit.runner.Description;
 import org.junit.runner.notification.RunNotifier;
 
-final class Spec implements Child {
+final class Spec implements Child, Atomic {
```
ashleyfrieze: A fine Blondie album

src/main/java/com/greghaskins/spectrum/Spectrum.java
```diff
@@ -55,7 +55,7 @@
    *
    */
   static void compositeSpec(final String context, final com.greghaskins.spectrum.Block block) {
-    final Suite suite = getCurrentSuiteBeingDeclared().addAbortingSuite(context);
+    final Suite suite = getCurrentSuiteBeingDeclared().addCompositeTest(context);
```
ashleyfrieze: clearer neater name

src/main/java/com/greghaskins/spectrum/Suite.java
```diff
@@ -46,31 +55,41 @@ static Suite rootSuite(final Description description) {
...
 
-  Suite addSuite(final String name, final ChildRunner childRunner) {
-    final Suite suite = new Suite(Description.createSuiteDescription(name), this, childRunner);
+  void completeSuiteDefinitionAndAdd(final Suite suite) {
     suite.beforeAll.addBlock(this.beforeAll);
     suite.beforeEach.addBlock(this.beforeEach);
     suite.afterEach.addBlock(this.afterEach);
+    rulesToApply.stream()
+        .map(RuleContextScope::nextGeneration)
+        .filter(scope -> scope != null)
```
ashleyfrieze: Optional won't work here - something to do with type erasure, I think.

src/main/java/com/greghaskins/spectrum/Suite.java
```diff
@@ -163,14 +191,34 @@ private void runChildren(final RunNotifier notifier) {
...
+  protected void runChild(final Child child, final RunNotifier notifier) {
     if (this.focusedChildren.isEmpty() || this.focusedChildren.contains(child)) {
-      child.run(notifier);
+      NotifyingBlock.wrap(
+          () -> runWithRules(rulesToApply.stream()
+              .filter(RuleContextScope::appliesToThisGeneration)
+              .map(RuleContextScope::get)
+              .collect(Collectors.toList()),
+              child, () -> runChildWithRules(child, notifier), getDescription()))
+          .run(getDescription(), notifier);
```
ashleyfrieze: This duplication needs removing before completion.

src/main/java/com/greghaskins/spectrum/internal/junit/StubJUnitFrameworkMethod.java
```diff
@@ -0,0 +1,35 @@
+package com.greghaskins.spectrum.internal.junit;
+
+import org.junit.runners.model.FrameworkMethod;
+
+/**
+ * Provide an framework method for JUnit to use as a stub for its method rules.
+ */
+public interface StubJUnitFrameworkMethod {
+  class Stub {
+    public void method() {}
```
ashleyfrieze: This is the so-called test method JUnit thinks is under test!

ashleyfrieze:
Still to do:

* Reconsider how the rule generations stuff works
* Make it possible for the test class/test object to run rules anyway!
* Make it possible to run rules against a class, but specify which annotated test method within it should be the rules' target - to allow for
```java
@DirtiesContext
@Test
public void testProcessWhichDirtiesAppCtx() {
    // some logic that results in the Spring container being dirtied
}
```

ashleyfrieze:
The class blocks of the unique rule class blocks should all be run once before the first actual test.

ashleyfrieze:
* Class rules now run along with `@BeforeClass` & `@AfterClass`
* The test object - i.e. the object that Spectrum created - can now host rules and things that rules govern.

src/main/java/com/greghaskins/spectrum/ConstructorBlock.java
ashleyfrieze: Yeah - I made that. It would be terrible in a parallel running scenario... potentially!

ashleyfrieze:
Now rebased to include #71. There may be a mix of metaphors in this merge, worth checking @greghaskins

src/main/java/com/greghaskins/spectrum/Spectrum.java
```diff
@@ -304,6 +329,22 @@ public static void aroundAll(ThrowingConsumer<com.greghaskins.spectrum.Block> co
...
+   * Insert a hook into the current level of definition.
+   * @param hook to insert
+   * @param appliesTo the {@link com.greghaskins.spectrum.model.HookContext.AppliesTo} indicating
+   *                  where the hook is run
+   * @param precedence the importance of the hook compared to others
+   */
+  public static void addHook(final Hook hook, final HookContext.AppliesTo appliesTo,
+      final HookContext.Precedence precedence) {
+    addHook(new HookContext(hook, getDepth(), appliesTo, precedence));
+  }
```
ashleyfrieze: I held off a public one of these until we needed it. We did here.

src/main/java/com/greghaskins/spectrum/model/Hooks.java
```diff
@@ -75,29 +71,33 @@ public void runAround(final Description description, final RunNotifier notifier,
...
+    return (description, notifier, innerBlock) -> {
+      hooksRememberedToRunTheInner.set(true);
+      innerBlock.run();
+    };
   }
 
+  private Hook wrap(final Hook inner, final HookContext outer) {
+    return (description, notifier, block) -> outer.getHook().accept(description, notifier,
+        wrapWithReporting(description, notifier,
+            () -> inner.accept(description, notifier, block)));
```
ashleyfrieze: Moving away from the original `ThrowingConsumer` to this has enabled the reporting to see the real exceptions.

src/test/java/specs/FixturesSpec.java
```diff
@@ -488,7 +488,7 @@
           it("report each error", () -> {
             assertThat(result.get().getFailureCount(), is(2));
 
-            assertThat(failureMessages.get(), hasItem("java.lang.Exception: boom 1"));
+            assertThat(failureMessages.get(), hasItem("boom 1"));
```
ashleyfrieze: Unfettered message.

ashleyfrieze:
Still to do:

- Prove that Spring JUnit rule actually works!
- Account for applying rules using a specific test method - i.e. one provided in the mix-in because it's got some annotations that a JUnit rule needs.

ashleyfrieze:
Spring.

Done!

Still to do:

- Account for applying rules using a specific test method - i.e. one provided in the mix-in because it's got some annotations that a JUnit rule needs.

e.g.

```java
	@Test
	@DirtiesContext(methodMode = DirtiesContext.MethodMode.AFTER_METHOD)
	public void willDoSomethingDestructiveToSpring() throws Exception {
		myBean.deleteAll();
        }
```

greghaskins:
Lots of really good work here. I really like that we have tested, working examples with Mockito and Spring, which are two of the biggest use-cases for this feature.

A few final things to resolve before merging:

1. _Minor:_ I'm seeing a bunch of compiler warnings about raw types and unchecked conversions. Is there any way to resolve those?
2. I'm really having a hard time understanding the difference between `applyRules` and `applyRulesHere`, and when you would use each. I can't follow what's going on in the example with `TemporaryFile`. It's kind of like `beforeEach` vs `beforeAll`, but not exactly? I guess this may also stem from me not fully understanding hooks yet, and the difference between `ATOMIC_ONLY` and `EACH_CHILD`.
    + Maybe they need clearer naming? Or documentation? Or examples?
3. I'm concerned that the "magic mode" (where Spectrum automatically applies any rules on the root level) will cause more confusion/questions than we'd like. On the surface it looks like you can do anything normal JUnit can do, but that's not quite correct. I'm guessing that misunderstanding is what happened [here](https://github.com/greghaskins/spectrum/issues/15#issuecomment-265604361), where Peter assumed it would "just work," and that others will have the same misunderstanding. I think we have a few options, with some combination of:
    + Add big warning stickers to the docs that say what you can and cannot do with with hooks and at each level of declaration. For instance, `@ClassRule` will not work on the test class itself (I think), so that needs to be a mixin. And magic rules can only be `EACH_CHILD` and not `ATOMIC_ONLY`, which means something like the difference between `applyRules` instead of `applyRulesHere` (and the improved clarity/explanation would be required here too).
    + Make the call to `applyRules(this)` explicit rather than implicit. That will mean people have to opt-in and are therefore more likely read said docs. This could just be a single line they add before any `describe` or `feature` that has the same effect as [this one](https://github.com/ashleyfrieze/spectrum/blob/cc5f4c1d7535e008bc3f416faac13a5ec751f043/src/main/java/com/greghaskins/spectrum/Spectrum.java#L453). If it is `applyRules(this)` then that also very clearly implies that you can only do instance-level stuff, since you're passing `this` and not a `Class<?>` object.
    + Or, as an alternative, we could have a `@RunWith(SpectrumWithRules.class)` variant of the runner that adds this magic. Just a different way to opt-in, but might allow us to support rules even better, since it could hypothetically work with `@ClassRule`.
    + We _could_ restrict the API to mixin-only, where we can make better guarantees about compatibility. I'm not saying we have to go that way, but it would certainly help clarify what is going on. Unfortunately, with the way other tools have tied themselves to `BlockJUnit4Runner`, we can't really just use JUnit ["as nature intended"](https://github.com/greghaskins/spectrum/pull/56/files#r91197132) in either format, because "nature" intended us to use `BlockJUnit4Runner` and didn't have things like Spectrum in mind. ☹️
4. Let's move the API entry points for JUnit rules out of `Spectrum.class` and into a separate interface and/or package. Even if the goal of decoupling from JUnit4 is not the immediate priority, at little healthy separation will make things easier down the line. 
5. The JUnit project has a convention of using `experimental` packages for features that are still baking. Gradle does something similar by marking features explicitly as ["incubating"](https://docs.gradle.org/current/userguide/feature_lifecycle.html) as opposed to "public." I'd be more willing to merge an evolving API if we could segregate and document it somehow like they do. I think we can get better real-world feedback by releasing this now and evolving it, rather than trying to lock down a "perfect" public API. Since this would be an "incubating" feature, that's another reason to make it opt-in instead of automatic.

As with #77, I'll probably make some minor post-merge changes as part of organizing the public API. I could even do some of the items above (particularly the last two) after merging, if that makes sense.

ashleyfrieze:
1. Type safety - resolved
2. All/Here/Each/Every... it's not great is it!

The junit stuff wants to work in a kind of "for each atomic" way. This means that, recursively down the tree, it will apply to `it`s and things like them. However, there's another approach. It MIGHT be better to apply the rules at the level of a particular suite. Within that suite, the rules run one for each immediate child, but don't propagate down to each child of each child.

The automagic JUnit stuff, to work the way JUnit users might expect, should be `each atomic` as should the default `applyRules` or whatever it's called. However, consider a situation where the rules set up and tear down a large eco system. We want to do that as few times as possible, and only repeat it if perhaps it's become invalidated. With this in mind you'd declare it at the suite level, and then have subsuites to reuse it.

If I'm not mistaken, `let` works at the suite level. Children of the children of the suite get the same object within `let`.

I think the name `applyRules` is not helping. I'd love to find a fitting alternative.

`Supplier<Mixin> = ...`

- `junitRulesAroundEach(Mixin.class)` / `junitRulesAroundAll(Mixin.class)`
- `junitRules(Mixin.class).forEachSpec()` / `junitRules(Mixin.class).forEachImmediateChild()`
- `applyRulesRecursively(Mixin.class)` / `applyRulesLocally(Mixin.class)`
- `junitRules(Mixin.class)` & `junitRules(Mixin.class, RECURSIVE)` / `junitRules(Mixin.class, IMMEDIATE_CHILDREN)`
- `rulesMixin(Mixin.class)` / `localRulesMixin(Mixin.class)`

3. I had a debate with myself about magic mode. I kind of expect it might end up being easier to agree on `SpectrumWithRules` but I'd like to make a case for signposting the JavaDoc and living with it being on by default. Opting out is very easy - namely just don't put any `@Rule` variants in your test class...

Here's what specifically went wrong for the DropWizard rules. I've never seen this happen before by the way, so I reckon this is a weak use case - i.e. <1%.

The DropWizard rule requires you to initialise the instance rule variable using the static rule variable. JUnit can guarantee that the instance variable won't be instantiated before the `@ClassRule` is run, because JUnit doesn't create any instance of the test object until a test starts. Spectrum MUST create an instance of the test object and only does that once.

This means there are two easily worked around limitations of Spectrum automagic;

- You can't instantiate an instance rule object from the static one
- You can't use rules which assume that there's a fresh instance of the test object every time

In both cases, you'd try, fail and drop to `applyRules` or whichever is the closest.

In 95%+ of cases, you can probably just use JUnit rules out of the box as JUnit was intended. It's a real shame to force people to type extra and people assume you can just use the rules.

While I appreciate the point about BlockJUnitRunner and Spectrum being different, the ONLY sticking point is actually the new object instance thing, and that's not actually such a big deal in practice.

4. Moving the interface - will do - let's just agree the names of things. Perhaps there's a JUnit class/interface in the main package we can put the rules functions into.
5. Use of `experimental` - we could do that.... I'm inclined to go for broke and just put this in the main, but why don't you reorganise it to suit yourself after the PR?

ashleyfrieze:
A thought that's been on my mind is the way that `EACH_CHILD` and `ATOMIC` propagation of hooks works, and thus how the two junit rules variants work.

It's the difference between shallow and deep propagation. As far as I understand it, `let` uses shallow propagation - the children of children of where a `let` is created get the same object.

Conversely, `beforeEach` uses deep propagation. Each child of a child of a child gets `beforeEach` called.

Now, it's possible that you think `let` should also be deep.

greghaskins:
Looking at this again, I'm actually willing to merge it mostly unchanged.

We definitely still have some confusion about the difference between `applyRules` and `applyRulesHere`. To me, the meaning behind `applyRules` is clear, but becomes muddy when compared against the `applyRulesHere` variant. JUnit's class rules are meant to be re-used between atomic tests, while method rules are meant to be fresh for each atomic test. With `applyRulesHere` we get a kind of middle ground where method rules actually do get re-used between some tests, but not all tests. I may have even gotten that wrong; and the difficulty explaining what it does is leading to difficulty naming it.

After looking at it for a while, I think I do understand the difference in propagation you're describing. The trouble, however, is the rules already describe how propagation should work (class vs method), and we're layering another system on top of that.

I propose just removing `applyRulesHere`. There may be scenarios where you have deep nesting and want to get fancy with what Spring (or whatever) re-initializes, but I don't think that's a particularly common case. Further, that would be something well beyond what "normal" JUnit could even do, which would be cool but not a deal breaker if we don't have it. 

If we take that out, I'm comfortable merging the rest as-is.

---

After this I do plan to re-organize some packages and such. There's probably a home for all the global JUnit functions in a common place, based on where that shakes out. I'll probably also touch up the documentation and maybe tweak a few tests to add clarity. 

There will of course be more significant changes as I push to decouple from JUnit (longer term), but I can't say exactly what that would look like yet. I'm thinking either `2.0` or `3.0` will make a significant break from the current packaging and API.

ashleyfrieze:
@greghaskins - the more I think about my reasoning for `applyRulesHere` the more I think it's based on instinct, not specific use case. I'll pull it out of the code. If we can't find a need for it, then it's just confusing. Can we rename `applyRules`? `junitMixin` instead - or `mixInJunit` - since that's really what we're doing.

If you have some minor clean-up glitches on your mind, I'm happy to tackle them a bit too. I appreciate you will want to do some general post-merge curation of your own. In my team, we use Pull Requests for Code Review and people give line-by-line feedback where it's needed, so I'm happy to receive per-line _could you play with that a bit_s if you want me to do the leg-work.

I'll ping this thread when the `applyRulesHere` is vanquished.

I'd _really_ like to keep the JUnit automagic that's also in this MR, and I notice you're not asking me to remove it. From the current docs, is there anything you'd like me to add regarding the caveats. I'm thinking that writing a separate article - linked from the `README.md` might might sense in discussing the pros and cons, similarities and differences, of each variant of the JUnit rules implementation. What do you think?

greghaskins:
I like the sound of `junitMixin`, let's go with that. 

There were only a couple small things I saw, I can add those as line comments here in the PR. None of them were deal-breakers, IMHO. More along the lines of "maybe play with this a bit" as you said.

I'm game to keep the automagic in and on by default. I think as we split out the JUnit specific pieces, that will make obvious sense for anyone using the JUnit runner (which is today the only runner).

Splitting out the details about how this works definitely seems like a good move. That would be a great first step toward the broader goal of better, proper docs instead of just a giant `README`. I might flip the order of presentation - to have people try automagic first, and falling back to `junitMixin` if that doesn't work. No need to bombard them with complexity for the 95% case where it will probably just work out of the box.

src/main/java/com/greghaskins/spectrum/Spectrum.java
```diff
@@ -371,7 +414,16 @@ private static Suite getCurrentSuiteBeingDeclared() {
...
   }
 
-  private static void addHook(HookContext hook) {
-    getCurrentSuiteBeingDeclared().addHook(hook);
+  /**
+   * If the definition is at the right level and the block is the right sort
+   * of block, then this is might be an opportunity to hook in JUnit rules.
+   * @param definitionBlock the block used to make this suite
+   */
+  private static void addRootJUnitHook(com.greghaskins.spectrum.Block definitionBlock) {
```
greghaskins: Maybe play with this implementation a little bit. It seems like we know declaratively when we are in the root block (in the `Spectrum` constructor), so the `if` statements and type cast should in theory not be needed.

src/main/java/com/greghaskins/spectrum/model/Hooks.java
```diff
@@ -3,19 +3,15 @@
...
 
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.LinkedList;
-import java.util.List;
 import java.util.function.Predicate;
-import java.util.stream.Collectors;
 
 /**
  * Collection of hooks. It is a linked list, but provides some helpers for
```
greghaskins: I get a compiler warning about `serialVerisionUID` on the `Hooks` class:

> The serializable class Hooks does not declare a static final serialVersionUID field of type long	Hooks.java	/spectrum/src/main/java/com/greghaskins/spectrum/model	     line 20

src/main/java/com/greghaskins/spectrum/internal/junit/Rules.java
```diff
@@ -0,0 +1,57 @@
...
+
+    return context;
+  }
+
+  /**
+   * Add a test object junit.rule set.
+   * @param object the JUnit test object - this will never be recreated
+   */
+  static void applyRules(Object object) {
+    RuleContext context = new RuleContext<>(object);
```
greghaskins: I get a compiler warning about raw types here:

> RuleContext is a raw type. References to generic type RuleContext<T> should be parameterized	Rules.java	/spectrum/src/main/java/com/greghaskins/spectrum/internal/junit	line 33

Maybe add `@SuppressWarnings("rawtypes")` to the local variable?

src/test/java/specs/JUnitRuleExample.java
```diff
@@ -0,0 +1,90 @@
...
+@RunWith(Spectrum.class)
+public class JUnitRuleExample {
+  public static class TempFolderRule {
+    @Rule
+    public TemporaryFolder tempFolderRule = new TemporaryFolder();
+  }
+
+  {
+    final Set<File> filesSeen = new HashSet<>();
+    describe("A parent spec", () -> {
```
greghaskins: This suite was a little hard to follow the first couple times through reading it. Maybe the assertions could be more direct, something like `assertTrue(theTempFolder.exists())` and/or  `assertThat(filesSeen, not(hasItem(theTempFolder)))`.

src/main/java/com/greghaskins/spectrum/Spectrum.java
```diff
@@ -223,6 +224,30 @@ public static Configuration configure() {
 
 
   /**
+   * Implements a rules context within the current suite using the given rules class as a mix-in.
+   * These rules will cascade down and be applied at the level of specs or atomic specs.
+   * @param rulesClass type of object to create and apply rules to for each spec.
+   * @param <T> type of the object
+   * @return a supplier of the rules object
+   */
+  public static <T> Supplier<T> applyRules(final Class<T> rulesClass) {
```
greghaskins: As discussed, we'll rename this one for better clarity. I'm still not sure where the best place for this entry point is, though. Not the worst if it's in root `Spectrum`, but as this is something I'll be marking as both experimental and JUnit-specific, it might make sense just letting people use `Rules.applyRules` directly, or maybe something like `JUnit.applyRules`. I'm flexible though.

ashleyfrieze: Renamed `junitMixin` - it shall stay here because the `junit` package is presently `internal` and should stay that way. I think `Spectrum` is quite a good Facade, but I appreciate you'll want to restructure and move packages about, so you'll have more freedom if I leave this alone.

src/main/java/com/greghaskins/spectrum/Spectrum.java
ashleyfrieze: Yeah - this could be done differently. So I changed it.

src/main/java/com/greghaskins/spectrum/internal/junit/Rules.java
ashleyfrieze: I went further and just propagated generic types down into the code - it turns out that the type things happen to be can be captured, making this notionally type safe.

I say notionally, since it kind of was, and it kind of doesn't matter, and it's kind of all just `Object` :)

src/main/java/com/greghaskins/spectrum/model/Hooks.java
ashleyfrieze: Ok. I've put one in. I don't see that warning, but it's no big deal.

ashleyfrieze:
I've address the comments. Please read my comments and please re-review.

src/test/java/specs/JUnitRuleExample.java
ashleyfrieze: With less to test, this was a bit clearer. However, I've done some renames of everything and I think it's probably better...

PR #75, opened 2016-12-02
Add `context` as alias for `describe`, similar to RSpec and Ginkgo

### LINKED ISSUE #63: Support `context` keyword from Ginkgo?
http://onsi.github.io/ginkgo/ has level of abstraction between `describe` and `it` in the form of `context`. This would be easy enough to add in the same way that the `GherkinSyntax` was added. Worth doing?

ashleyfrieze:
Resolves #63


greghaskins:
Looks good overall. There's actually nothing Ginkgo-specific about `context` – it's available in RSpec, Mocha, Kiwi, and others too. So, two tweaks:
1. Let's move these functions to the main `Spectrum` class right alongside `describe` and friends.
2. They don't need a whole section in the README, we can document them in the same way as `describe`, `it`, and such. That probably just means moving the examples to `ExampleSpecs` and calling it a day.

ashleyfrieze:
This change is way way smaller than it used to be., @greghaskins I reckon this could be an easy merge for you..., whaddya think?

Issue #76, opened 2016-12-03
Consolidation refactoring - before/around/let - the grand unified theory of hooks


ashleyfrieze:
There are several constructs within Spectrum which are essentially hooks that run before/after or both before and after execution blocks. The `let` supplier is refreshed on a test by test basis. The `beforeAll` is run around a set of items. The `beforeEach` runs before individual specs.

All of these could be consolidated into a hook plugin architecture. We keep the user interface and just modify the internals so that there's a chain of hooks at each level of the hierarchy. Those hooks may or may not be connected with Supplier references that the test definition has kept. The hooks may or may not run at the level of a suite or down at the individual "test". For this to work, the concept of `Atomic` already incubating on #56 will be useful.

The hooks will also be exposed with functions `hook` and `localHook` or similar, for ANY external plugin to be possible. This means the JUnit rules implementation of #56 can become relegated to a plugin.

Hopefully a lot of code will go away in `Suite` with this change, as there'll only really be one implementation of lots of things suite does as a chain.

ashleyfrieze:
@greghaskins - this can now be closed, I think.

greghaskins:
Closed by #77

Issue #73, opened 2016-12-02
aroundAll runs even when all specs are ignored


greghaskins:
This is a low-priority one, but should probably get addressed at some point. At the moment, there isn't a simple implementation fix, because `Suite` objects don't know which of their children are ignored or not. I think the underlying implementation will evolve to be based more around tags anyway; fixing this issue will be simpler after that.

For reference, this is the failing test:

```java
describe("a suite with only ignored specs", () -> {

  it("should not run aroundAll", () -> {
    ArrayList<String> steps = new ArrayList<>();
    SpectrumHelper.run(() -> {

      aroundAll(block -> {
        steps.add("aroundAll");
        block.run();
      });

      xit("foo", () -> {
      });

    });

    assertThat(steps, is(empty()));
  });

});
```

ashleyfrieze:
The suite should be able to tell if its children are ignored. The question is, I guess, whether it needs to be testing for ignored of direct descendants, or the whole tree.

Just add `isIgnored` to `Child` and then 

```java
// suite
boolean areChildrenIgnored() {
    return children.stream().filter(child -> !child.isIgnored()).findFirst().isPresent();
}
``

ashleyfrieze:
This will be fixed by #56 which would otherwise be doing big class-rule things with totally ignored suites.

greghaskins:
@ashleyfrieze we should check if this bug still applies, or whether it was fixed by #56 .

ashleyfrieze:
````java
@RunWith(Spectrum.class)
public class AroundAllSpec {{
	describe("A suite with an around all", () -> {
		aroundAll(block -> {
			System.out.println("Before running");
			block.run();
			System.out.println("after running");
		});
		
		xit("should not run", () -> {});
	});
}}
````

The above produces

````
Before running

Test ignored.
after running
````

ashleyfrieze:
So the bug still applies

ashleyfrieze:
I think this may have been fixed by #56 

````java
      describe("a suite with only ignored specs", () -> {

        it("should not run aroundAll", () -> {
          ArrayList<String> steps = new ArrayList<>();
          SpectrumHelper.run(() -> {

            aroundAll(block -> {
              steps.add("aroundAll");
              block.run();
            });

            xit("foo", () -> {
            });

          });

          assertThat(steps, is(empty()));
        });

      });
````

[...]

Issue #67, opened 2016-11-30
Executing tests in parallel would be impaired by `synchronized` block in `Spectrum`


ashleyfrieze:
In a situation where something like JUnit's `ParallelComputer` were used, or where other test runners were parallelising the construction of test classes run with the `Spectrum` runner, the fact that the test class construction happens inside a synchronized block will constrain the execution time.

This is relatively easy to fix. It's in a synchronized block to protect a stack from being bombarded by different threads. By introducing a thread-independent singleton into the mix, this will become thread-safe. This is what I propose to do.

ashleyfrieze:
@greghaskins - this was resolved by #77

ashleyfrieze:
I'm going to close this, since it's a detail, rather than an issue.

Issue #61, opened 2016-11-30
Run specs or suites in Parallel


ashleyfrieze:
I intend to build some Acceptance tests for a batch processing system. These generally have the form of:

```gherkin
Given we drop file X into the input
When the system has a result in the output
Then it is a good one
```

As you might imagine, each test uploading files or waiting for propagation time is very slow in terms of elapsed time, even though the system under test could receive all inputs at once and run them all in parallel. While we could write the whole test as a tangle of parallel concerns:

```gherkin
Given we drop file X into the input
And we drop file Y into the input
When the system has results in the output for X AND Y
Then X is a good one
And Y is a good one
```

The best option would be to write the tests to run in parallel. Spectrum style it might look like this:

```java
feature("unconflicting use cases that can share the system", with(parallelRunning(), () -> {
   scenario("the X files", () -> {
       given("we drop the file into the input", () -> {
          system.drop("X.file");
       });
      when("the system has results", () -> {
          system.waitFor("X.result");
      });
      then("the result is good", () -> {
          validator.check("X.result", "X.expectedResult");
      });
   }); 
   
   // this scenario is expected to run in parallel with the other
   scenario("the y files", () -> {
       given(...);
       when(...);
       then(...);
   });
});
```

This will play havoc with the output to Eclipse or IntelliJ's runner, though potentially a RunNotifier proxy could be put in the middle to avoid that.

ashleyfrieze:
@greghaskins if I distil the above into the following question, how would you react?

What if _execute all specs in parallel_ could be done by a `PreCondition` like `parallelRunning` or `runInParallel`? Would that be in keeping with your design expectations?

PR #77, opened 2016-12-04
Grand unified theory of hooks

ashleyfrieze:
Still WIP at the moment - demonstrated purely by a replacement of the `let` function.

I've reorganised the package structure, and I'm in the process of killing off all arounds/befores etc. First target - the humble `let`.

Code coverage will not be 100% yet as there are unused features.

Intended as a fix for #76 and #73. Eventually. Intended as a new platform on which to build #56


ashleyfrieze:
Highlights:

- Large package structure change of the internal classes. There's now an `internal` and a `model` package into which detail has been stuck to hide
- All the per-use-case implementation of befores, around, afters etc has been turned into a general purpose `hook` mechanism.
  - to get the tests to pass, though, this mechanism has been twisted a bit
  - hooks are now classified by where they run, rather than what they are
  - helpers for before/after and around are there
  - the way is paved for hooks that are suppliers, for which 'let' is an example
  - JUnit rules would be an example and could now be implemented independently of `Spectrum` as just another plugin hook
- The nature of notifyingblocks and other general purpose error handling has modified the error messages slightly
- The way that hooks can stop execution means that some befores and afters don't happen the way they used to - what they do seems logical enough to me, though.

src/main/java/com/greghaskins/spectrum/AbstractSupplyingHook.java
```diff
@@ -0,0 +1,47 @@
...
+import com.greghaskins.spectrum.model.Singleton;
+
+/**
+ * A base class for supplying hooks to use. Override before or after. Return the singleton
+ * value from the before method.
+ * You can use this to write any plugin which needs to make a value visible to the specs.
+ * This is not the only way to achieve that - you can also build from {@link SupplyingHook}
+ * but this captures the template for a complex hook.
+ */
+public class AbstractSupplyingHook<T> extends Singleton<T> implements SupplyingHook<T> {
```
ashleyfrieze: This would do both `let` and `junitTestClass`

src/main/java/com/greghaskins/spectrum/AbstractSupplyingHook.java
```diff
@@ -0,0 +1,47 @@
...
+  public void acceptOrThrow(Block block) throws Throwable {
+    set(before());
+    block.run();
+    after();
+    clear();
+  }
+
+  @Override
+  public T get() {
+    assertSpectrumInTestMode();
```
ashleyfrieze: Borrowed from `let` not a bad thing to have around.

src/main/java/com/greghaskins/spectrum/Configuration.java
```diff
@@ -1,5 +1,9 @@
 package com.greghaskins.spectrum;
 
+/**
+ * Allows the injection of suite configuration during test definition. A wrapper for the
+ * suite which exposes configurables.
+ */
```
ashleyfrieze: Checkstyle forced me to write Javadoc as items stopped being internal when they switched packages.

src/main/java/com/greghaskins/spectrum/Hook.java
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * A hook allows you to inject functionality before and/or after a {@link Block}.
+ * Just implement the {@link ThrowingConsumer#acceptOrThrow(Object)} method and
+ * call {@link Block#run()} within your implementation.
+ * If your hook is going to provide an object to the running test, then implement
+ * {@link SupplyingHook} or subclass {@link AbstractSupplyingHook}.
+ */
+public interface Hook extends ThrowingConsumer<Block> {
+}
```
ashleyfrieze: I really want to modify `aroundXXX` to use this rather than `ThrowingConsumer` - we probably could, but for now I've left an adapter in place.

src/main/java/com/greghaskins/spectrum/NotifyingBlock.java
```diff
@@ -1,26 +0,0 @@
-package com.greghaskins.spectrum;
```
ashleyfrieze: not dead -> moved.

src/main/java/com/greghaskins/spectrum/Spectrum.java
```diff
@@ -303,7 +296,7 @@ default T get() {
    * @param consumer to run each spec block
    */
   public static void aroundEach(ThrowingConsumer<com.greghaskins.spectrum.Block> consumer) {
-    getCurrentSuiteBeingDeclared().aroundEach(consumer);
+    getCurrentSuiteBeingDeclared().aroundEach(consumer::acceptOrThrow);
```
ashleyfrieze: Adapted!

src/main/java/com/greghaskins/spectrum/Spectrum.java
```diff
@@ -271,27 +282,9 @@ public static void afterAll(final com.greghaskins.spectrum.Block block) {
...
-    default T get() {
-      try {
-        return getOrThrow();
-      } catch (final RuntimeException | Error unchecked) {
-        throw unchecked;
-      } catch (final Throwable checked) {
-        throw new RuntimeException(checked);
-      }
-    }
+  public interface ThrowingSupplier<T> extends com.greghaskins.spectrum.ThrowingSupplier<T> {
```
ashleyfrieze: Kept for posterity...

src/main/java/com/greghaskins/spectrum/Suite.java
```diff
@@ -9,18 +23,13 @@
...
-  private final SetupBlock beforeAll = new SetupBlock();
-  private final TeardownBlock afterAll = new TeardownBlock();
-
-  private final SetupBlock beforeEach = new SetupBlock();
-  private final TeardownBlock afterEach = new TeardownBlock();
+class Suite implements Parent, Child {
+  // the hooks - they will be turned into a chain of responsibility
+  // so the first one will be executed last as the chain is built up
+  // from first to last.
+  private Hooks hooks = new Hooks();
```
ashleyfrieze: So much smaller.

src/main/java/com/greghaskins/spectrum/Suite.java
```diff
@@ -97,35 +110,7 @@ private Child createSpec(final String name, final Block block) {
...
-          this.afterEach.run(description, notifier);
-        });
-
-
-        if (!blockWasRun.get()) {
-          throw new RuntimeException("aroundEach did not run the block");
-        }
-
-      }).run(description, notifier);
-    };
```
ashleyfrieze: Lots of code goes bye bye...

src/main/java/com/greghaskins/spectrum/Suite.java
```diff
@@ -213,34 +232,33 @@ public void run(final RunNotifier notifier) {
...
+      runChildren(notifier);
+    });
   }
 
   private void runChildren(final RunNotifier notifier) {
     this.childRunner.runChildren(this, notifier);
   }
 
-  private void runChild(final Child child, final RunNotifier notifier) {
+  protected void runChild(final Child child, final RunNotifier notifier) {
```
ashleyfrieze: To enable the child to be run from the composite test subclass

ashleyfrieze:
Booyah! Coverage 100%!

src/main/java/com/greghaskins/spectrum/Hook.java
greghaskins: We haven't done a release with `ThrowingConsumer`, so I wouldn't be too worried about protecting that API.

src/test/java/specs/FixturesSpec.java
```diff
@@ -483,11 +484,10 @@
               .map((failure) -> failure.getMessage())
               .collect(Collectors.toList()));
 
-          it("report each error", () -> {
-            assertThat(result.get().getFailureCount(), is(2));
+          it("report the error only once", () -> {
```
greghaskins: Note: Let's not change this until we come to a decision on the other "to discuss" items.

src/main/java/com/greghaskins/spectrum/model/TaggingState.java
```diff
@@ -10,17 +12,17 @@
 /**
  * Represents the state of tagging for Spectrum - what it presently means.
  */
-class TaggingState {
+public class TaggingState {
```
greghaskins: If this is getting promoted to the public API, then we might need to consider it's name. At first, I thought `TaggingState` captured the tags for a spec, not the filter of which tags to run and which to ignore. Perhaps `FilterCriteria` or something like that would be more appropriate.

src/test/java/specs/AroundSpecs.java
```diff
@@ -71,8 +71,7 @@
 
         assertThat(result.getFailureCount(), is(1));
         assertThat(result.getFailures().get(0), is(failure("first spec",
-            RuntimeException.class, "aroundEach did not run the block")));
-
+            RuntimeException.class, "At least one of the test hooks did not run the test block.")));
```
greghaskins: With the generic hooks implementation, we certainly can't give as specific a message. Maybe we can at least include the location/trace to the offending hook?

src/test/java/specs/TaggedSpecs.java
```diff
@@ -1,12 +1,12 @@
 package specs;
 
-import static com.greghaskins.spectrum.PreConditionBlock.with;
-import static com.greghaskins.spectrum.PreConditions.Factory.tags;
 import static com.greghaskins.spectrum.Spectrum.beforeEach;
 import static com.greghaskins.spectrum.Spectrum.configure;
 import static com.greghaskins.spectrum.Spectrum.describe;
 import static com.greghaskins.spectrum.Spectrum.it;
 import static com.greghaskins.spectrum.Spectrum.let;
+import static com.greghaskins.spectrum.internal.PreConditionBlock.with;
```
greghaskins: Let's discuss package structure and naming for PreConditions. If we are going to generalize the API to support things like parallelization, timeouts, etc. then we might want a different term. It feels kind of like some sort of `Decorator` maybe?

src/main/java/com/greghaskins/spectrum/AbstractSupplyingHook.java
```diff
@@ -0,0 +1,45 @@
...
+   * Template method for a hook which supplies.
+   * @param block the inner block that will be run
+   * @throws Throwable on error
+   */
+  @Override
+  public void acceptOrThrow(Block block) throws Throwable {
+    set(before());
+    block.run();
+    after();
+    clear();
```
greghaskins: At a minimum, `clear()` should be in a `finally` block. Depending on the outcome of other discussions, `after()` may also need to be.

src/main/java/com/greghaskins/spectrum/Suite.java
```diff
@@ -139,19 +124,44 @@ private void addChild(final Child child) {
...
+   * @param hook to add
+   */
+  void addHook(final HookContext hook) {
+    this.hooks.addFirst(hook);
+  }
+
+  Hooks getHooksFor(final Child child) {
+    Hooks allHooks = parent.getInheritableHooks().plus(hooks);
+
+    return child instanceof Atomic ? allHooks.forAtomic() : allHooks.forNonAtomic();
```
greghaskins: Tiny nit: is there a way to do this without `instanceof`? Seems like a possible design smell. Maybe `Child` just has an `isAtomic()` method. `Spec` and `CompositeTest` would say `true`, while `Suite` would say `false`.

greghaskins:
@ashleyfrieze FYI I left several context-specific comments over in [https://github.com/greghaskins/spectrum/pull/77/files](). I'm not sure what GitHub will give you notifications about, I wanted to make sure you saw those.

Re: "to discuss". You've raised the question about whether Spectrum should run `after` even when a `before` fails, and whether to report all errors individually as failures. The quest to get that behavior "right" started with me running RSpec, Jasmine, and vanilla JUnit a lot, and then mimicking what they do to build out the first pass. Later we had #40 come in as a bug report, and in the interim fix PR (#41) we had a good discussion. Particularly, see my [comment here](https://github.com/greghaskins/spectrum/pull/41#issuecomment-238729178). The description of expected behaviors and the implementation to resolve it properly came in #45 . 

I would encourage you to read through those threads and share your thoughts here. I'll definitely acknowledge that matching those behaviors adds significant implementation complexity. It also keeps Spectrum lock-step in line with RSpec and friends. The RSpec team, in particular, has taken great care to define and document the behavior for error scenarios (see [their relish docs](https://www.relishapp.com/rspec/rspec-core/v/3-5/docs/hooks/before-and-after-hooks) and their [test suite](https://github.com/rspec/rspec-core/blob/0c8db72445c21156757c8f1069aeca3c8c6841ed/spec/integration/suite_hooks_errors_spec.rb#L86-L131)).

greghaskins:
One issue with this unified hooks implementation is that it treats `beforeEach` and `afterEach` with the same precedence as `aroundEach`. In RSpec, `around` hooks are guaranteed to run _around_ any `before`/`after` hooks. This is particularly noticeable when those hooks are declared in a slightly different order that we currently test them.

## RSpec example

```ruby
RSpec.describe "specs with hooks" do

  before(:each) do
    puts "before1"
  end

  around(:each) do |block|
    puts "around1 - pre"
    block.run
    puts "around1 - post"
  end

  after(:each) do
    puts "after1"
  end

  before(:each) do
    puts "before2"
  end

  around(:each) do |block|
    puts "around2 - pre"
    block.run
    puts "around2 - post"
  end

  it 'should run them' do
    puts "-SPEC-"
  end

  after(:each) do
    puts "after2"
  end

end
```

Output:

> around1 - pre
around2 - pre
before1
before2
-SPEC-
after2
after1
around2 - post
around1 - post

## Spectrum example (current implementation for #77)

```java
describe("specs with hooks", () -> {

  beforeEach(() -> System.out.println("before1"));

  aroundEach(block -> {
    System.out.println("around1 - pre");
    block.run();
    System.out.println("around1 - post");
  });

  afterEach(() -> System.out.println("after1"));

  beforeEach(() -> System.out.println("before2"));

  aroundEach(block -> {
    System.out.println("around2 - pre");
    block.run();
    System.out.println("around2 - post");
  });

  it("should run them", () -> System.out.println("-SPEC-"));

  afterEach(() -> System.out.println("after2"));
});
```

Output:
> before1
around1 - pre
before2
around2 - pre
-SPEC-
after2
around2 - post
after1
around1 - post

This seems to be solely driven by declaration order, rather than a semantic distinction between the hooks. I'm open to discussing whether we should care if it matches--you can guess my default stance.

For completeness, the same suite using the implementation on master (903d5a6) does match RSpec:

> around1 - pre
around2 - pre
before1
before2
-SPEC-
after2
after1
around2 - post
around1 - post

FWIW this gets even hairier when you throw nesting into the mix.

src/test/java/specs/FixturesSpec.java
ashleyfrieze: Totally - was a temp test change. Let's revert it.

src/main/java/com/greghaskins/spectrum/model/TaggingState.java
ashleyfrieze: `TaggingFilterCriteria`

src/test/java/specs/AroundSpecs.java
ashleyfrieze: Can you suggest how?

src/main/java/com/greghaskins/spectrum/model/Singleton.java
```diff
@@ -0,0 +1,26 @@
...
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
+
+/**
+ * The singleton pattern. An abstraction of a single object which can be created and
+ * destroyed. This is a twist on the usual singleton pattern, in that normally the
+ * singleton stays around indefinitely. This version is also thread safe.
+ */
+public class Singleton<T> implements Supplier<T> {
```
ashleyfrieze: My aim is to make it really thread safe in #61 for now, I'm just laying foundations.

I'm going to change the singleton stuff. You're right. It's got surprise in it. I'm going to claim that I am right to have chosen the name, since a Singleton is actually capable of abstracting the correct creation of an object and having more than one if you're doing things across threads.

In fact, though. The `Variable` needs to be threadsafe, and it seems like `let` and `Variable` ought to share more implementation. Watch this space.

src/main/java/com/greghaskins/spectrum/Suite.java
ashleyfrieze: There is a way to do this without instanceof. I'm going to say that it doesn't matter either way. Either we make a default method that returns false and override it to return true in Spec and CompositeTest, or we just use a tagging interface and query with instanceof.

For the avoidance of debate, I'll just change it.

ashleyfrieze:
I have an idea. I think I can fix the afters by sorting the hooks. I'll try it out! It will make some other yucky stuff go away too.

ashleyfrieze:
Now also addresses #67

src/test/java/specs/AroundSpecs.java
ashleyfrieze: Technically a hook could execute an it over multiple examples... so preventing multiple runs might be wrong...but this is not the easiest of frameworks to debug a failing test in. The simpler the hooks the better. 

Let's defer all of this for post merge.

src/main/java/com/greghaskins/spectrum/Hook.java
ashleyfrieze: Throwing Consumer can die then.

src/main/java/com/greghaskins/spectrum/AbstractSupplyingHook.java
ashleyfrieze: Done

ashleyfrieze:
In this increment, the hook contexts are sortable by importance. This enables prioritisation between types of hook. The `afterAll` tests are still not passing, which implies some other issue with how the code is executed. We're getting closer, but not quite there yet.

There's a runtime cost of the `sorted` function in `Hooks`. This may be cacheable. It's also fairly trivial.

I think this is pretty close to where it needs to be.

src/test/java/specs/FixturesSpec.java
```diff
@@ -113,11 +114,15 @@
...
+          // looking at the beforeEach statements, they ARE in this order
+          // to discuss
+          assertThat(numbers, contains(1, 3, 4, 2));
         });
 
         it("runs afterEach blocks from parent context", () -> {
-          assertThat(numbers, contains(1, 2, 3, 4));
+          // TODO - how is this testing the after each?
+          // just through the numbers not stacking up more?
+          assertThat(numbers, contains(1, 3, 4, 2));
```
greghaskins: Agreed, this is not a clear test at all. It would be better to test `afterEach` directly, similar to `beforeEach`.

src/test/java/specs/FixturesSpec.java
```diff
@@ -113,11 +114,15 @@
         });
 
         it("all run before each test in declaration order", () -> {
-          assertThat(numbers, contains(1, 2, 3, 4));
+          // looking at the beforeEach statements, they ARE in this order
+          // to discuss
+          assertThat(numbers, contains(1, 3, 4, 2));
```
greghaskins: The order should be 1, 2, 3, 4 here. That's to match the RSpec and Jasmine ordering. Child contexts implicitly depend on their parent (since you can access the parent's state), so you need to make sure the parent scope gets set up first before setting up the child scope. Same for teardown code; can't tear down the outer scope until the inner scope (that might rely on outer scope state) has cleaned up it's own stuff.

For a concrete example, consider an integration suite that hits a database. The outermost scope might begin a transaction in `beforeEach` and roll it back in `afterEach`. If the nested context does some database manipulation in its own `beforeEach`, that should run inside the wrapping transaction. If it does some cleanup or verification in an `afterEach`, that should happen before the transaction rolls back.

I made up another RSpec sample for reference, and to check my own sanity:

```ruby
RSpec.describe "specs with hooks" do

  before(:each) do
    puts "before1"
  end

  after(:each) do
    puts "after1"
  end

  before(:each) do
    puts "before2"
  end

  it 'something' do
    puts "-OUTER SPEC-"
  end

  describe "and nesting" do 

    before(:each) do 
      puts "before3"
    end

    it "something else" do
      puts "-INNER SPEC-"
    end

    after(:each) do 
      puts "after3"
    end

  end

  after(:each) do
    puts "after2"
  end

end
```
Output: 

> before1
before2
-OUTER SPEC-
after2
after1
before1
before2
before3
-INNER SPEC-
after3
after2
after1

src/main/java/com/greghaskins/spectrum/model/HookContext.java
```diff
@@ -0,0 +1,146 @@
...
+
+  @Override
+  public int compareTo(HookContext other) {
+    if (precedence == other.precedence) {
+      return Integer.compare(other.sequenceNumber, sequenceNumber);
+    }
+
+    return Integer.compare(precedence.getOrdering(), other.precedence.getOrdering());
+  }
+}
```
greghaskins: Very nice refactor here to add `Precedence` and `AppliesTo`. Quite clever!

At the end of (or after we merge) this PR, I'll take a shot at documenting the whole hook mechanism and API. Doing that will serve two purposes:
1. Said documentation will be helpful to users and plugin authors.
2. It will verify that I myself have a good understanding of what's going on.
3. It might reveal further tweaks to naming, etc. that would make comprehension easier before we go live with a release.

ashleyfrieze: Thanks. If I add depth hierarchy too, then the 1234/1423 test will pass.

src/test/java/specs/FixturesSpec.java
ashleyfrieze: Made a test which tests for what we're really testing for. That the list is refreshed each time.

src/main/java/com/greghaskins/spectrum/AbstractSupplyingHook.java
ashleyfrieze: And by borrowed, I mean stolen, given that `let` is now a hook!

src/main/java/com/greghaskins/spectrum/Hook.java
ashleyfrieze: `ThrowingSupplier` on the other hand, was the item in question and I've removed it.. then restored it.

src/main/java/com/greghaskins/spectrum/Spectrum.java
```diff
@@ -188,7 +205,8 @@ public static Configuration configure() {
    * @param block {@link com.greghaskins.spectrum.Block} to run once before each spec
    */
   public static void beforeEach(final com.greghaskins.spectrum.Block block) {
-    getCurrentSuiteBeingDeclared().beforeEach(block);
+    addHook(new HookContext(before(block), getDepth(),
+        HookContext.AppliesTo.ATOMIC_ONLY, HookContext.Precedence.LOCAL));
```
ashleyfrieze: all hookcontexts created by `Spectrum` which finds it is to provide the depth.

The Precedence names and application are arbitrary and "whatever I had to do to get the tests to work"

src/main/java/com/greghaskins/spectrum/Spectrum.java
```diff
@@ -314,11 +276,46 @@ public static void aroundEach(ThrowingConsumer<com.greghaskins.spectrum.Block> c
...
+   */
+  public static void assertSpectrumInTestMode() {
+    if (getCurrentSuiteBeingDeclared() != null) {
+      throw new IllegalStateException("Cannot use this statement in a suite declaration. "
+          + "It may only be used in the context of a running spec.");
+    }
+  }
 
-  private static final Deque<Suite> suiteStack = new ArrayDeque<>();
+  private static final Variable<Deque<Suite>> suiteStack = new Variable<>(ArrayDeque::new);
```
ashleyfrieze: It's a threadsafe `Deque` - I guess this is itself just a wrapper for `ThreadLocal` but it feels nice using a Spectrum construct.

src/main/java/com/greghaskins/spectrum/Suite.java
```diff
@@ -42,7 +43,8 @@
   }
 
   static Suite rootSuite(final Description description) {
```
ashleyfrieze: `rootSuite` was a song from Chitty Chitty Bang Bang.

src/test/java/given/a/spec/with/exception/in/aftereach/block/WhenRunningTheSpec.java
```diff
@@ -27,7 +27,8 @@ public void thereIsOneFailureForEachAffectedTest() throws Exception {
   @Test
   public void theFailureExplainsWhatHappened() throws Exception {
     assertThat(this.result.getFailures().get(0),
-        is(failure("a passing test", Fixture.SomeException.class, "kaboom")));
+        is(failure("a passing test", RuntimeException.class,
+            "given.a.spec.with.exception.in.aftereach.block.Fixture$SomeException: kaboom")));
```
ashleyfrieze: Changed purely down to the checked exception wrapping that now happens in `ThrowingConsumer`

src/test/java/given/a/spec/with/exception/in/beforeeach/block/and/aftereach/block/Fixture.java
```diff
@@ -13,11 +13,11 @@
         describe("an exploding beforeEach", () -> {
 
           beforeEach(() -> {
-            throw new SomeException("kaboom");
+            throw new SomeException("beforeEach went kaboom");
```
ashleyfrieze: Just made the message more diagnostic.

src/test/java/given/a/spec/with/exception/in/beforeeach/block/and/aftereach/block/Fixture.java
```diff
@@ -34,7 +34,7 @@
     return Spec.class;
   }
 
-  public static class SomeException extends Exception {
+  public static class SomeException extends RuntimeException {
```
ashleyfrieze: Changed to unchecked to preserve original message.

src/test/java/given/a/spec/with/exception/in/beforeeach/block/and/aftereach/block/WhenRunningTheSpec.java
```diff
@@ -28,9 +28,11 @@ public void thereAreTwoFailuresForEachAffectedTest() throws Exception {
   @Test
   public void theFailuresExplainWhatHappened() throws Exception {
     assertThat(this.result.getFailures().get(0),
-        is(failure("a failing test", Fixture.SomeException.class, "kaboom")));
+        is(failure("a failing test", Fixture.SomeException.class,
+            "beforeEach went kaboom")));
```
ashleyfrieze: Here's where the test uses the more specific error message - no tricks!

src/test/java/specs/FixturesSpec.java
```diff
@@ -113,11 +113,15 @@
...
         it("all run before each test in declaration order", () -> {
+          // looking at the beforeEach statements, they ARE in this order
+          // to discuss
           assertThat(numbers, contains(1, 2, 3, 4));
         });
 
         it("runs afterEach blocks from parent context", () -> {
-          assertThat(numbers, contains(1, 2, 3, 4));
+          // if afterEach block didn't call clear, then
+          // the numbers would be of size 8, not a fresh four
```
ashleyfrieze: Added narrative to explain what the test means.

src/test/java/specs/FixturesSpec.java
```diff
@@ -486,7 +490,7 @@
           it("report each error", () -> {
             assertThat(result.get().getFailureCount(), is(2));
 
-            assertThat(failureMessages.get(), hasItem("boom 1"));
+            assertThat(failureMessages.get(), hasItem("java.lang.Exception: boom 1"));
```
ashleyfrieze: The checked exception gets wrapped. This only happens in the innermost consumer. Outer layers are (and can be) more permissive of unchecked exceptions.

src/test/java/specs/FixturesSpec.java
```diff
@@ -502,12 +506,12 @@
           (result) -> () -> result.get()
               .getFailures()
               .stream()
-              .map((failure) -> failure.getMessage())
+              .map(Failure::getMessage)
```
ashleyfrieze: Function reference beats lambda.

src/test/java/specs/FixturesSpec.java
```diff
@@ -502,12 +506,12 @@
...
               .collect(Collectors.toList());
 
       describe("in beforeEach and afterEach", () -> {
 
-        final Supplier<List<String>> exceptionsThrown = let(() -> new ArrayList<>());
+        final Supplier<List<String>> exceptionsThrown = let(ArrayList::new);
```
ashleyfrieze: Function reference again.

src/test/java/specs/FixturesSpec.java
```diff
@@ -519,21 +523,21 @@
           describe("an explosive suite", () -> {
 
             beforeEach(() -> {
-              throw recordException.apply(new Exception("boom beforeEach 1"));
+              throw recordException.apply(new RuntimeException("boom beforeEach 1"));
```
ashleyfrieze: Runtime exceptions make the tests easier to read/write.

src/test/java/specs/FixturesSpec.java
greghaskins: Nice! I'm still getting used to all these cool Java 8 features.

greghaskins:
@ashleyfrieze Most excellent! A very nice overhaul of the codebase, with all original tests intact and passing!

I'm prepared to merge this PR as-is, with a few to-dos on my list before the next release. I think it makes sense to resolve these as separate PRs, since this one is big enough!

With all the features we've been adding, it feels like we're approaching a `2.0` milestone. With that in mind, I'd like to 
- Organize the public API as a whole
  - Further package structure overhaul (make more stuff `internal` and split out `syntax` or `dsl` package)
  - Move to `org.spectrumbdd` namespace for 2.0, keep `com.greghaskins.spectrum` for 1.0-compatibility, but `@Deprecated`
- Write more documentation in general, and for hooks specifically. This might lead to API tweaks for clarity and ease-of-use.

The above bits will probably affect this code, perhaps moving pieces around and other minor changes. If you're good with that, then I will hit the Merge button on #77.

ashleyfrieze:
Yeah. Merge merge merge!

I can redo my JUnit rules piece then. Awesome!

PR #88, opened 2017-02-01
Organize and cleanup new features before release (part 1)

greghaskins:
I'm working on a few things in this PR, and will update it as I go. 

In no particular order:

- [x] Put all public APIs in one main package for easier discovery and import. Many of them will be available directly under `Spectrum.class`. 
    - `com.greghaskins.spectrum.internal.ConfiguredBlock`
    - `com.greghaskins.spectrum.model.BlockConfiguration.Factory`
- [x] Move as many internal things to the `internal` packages as possible. This will clarify the real public API.
- [x] Update version numbers and deprecation notices in javadocs
- [ ] Make sure nothing `internal` is involved in the public API (either as entry point, parameter, or return value). Some of these will need to move to main package as a result.
- [ ] Resolve cyclic dependencies, at least involving the main package `com.greghaskins.spectrum` and it's classes.
- [ ] ~Reduce the logging noise from the test suite. Mainly the Spring stuff spits a bunch of `stderr` messages.~
- [ ] ~Determine which features should be marked and packaged as `spectrum.experimental`~


greghaskins:
@ashleyfrieze @richdouglasevans I'm partway through my work on this PR, and will be merging what I have so far. My goal is to release a new version as soon as feasible, so I've made some decisions to accelerate that.

1. I'm going to define the "Public API" to be all public methods on public classes/interfaces in `com.greghaskins.spectrum`, including their parameters and return types. I still have a few things to move around related to parameterized specs. I think I have stripped this down to the minimum it needs to be, but could use s second set of eyes.
2. I decided against having an `experimental` sub-package. Everything is just public or internal; less confusion that way.
2. For now, I'm going to keep the `addHook` API internal. There's still some refinement needed on the terminology and `HookContext` in particular, and I don't want that to delay a release of other valuable features we've got. Everything is still implemented as hooks internally, that extension point just isn't exposed externally yet.
3. I made a crude attempt to split the API for `BlockConfiguration` into the parts I felt should be public vs internal. I'm not 100% thrilled with how it turned out, but I've got a `BlockConfigurationChain` that essentially has the chainable `.focus().ignore().tags(...)` methods on it. I'm open to better ideas, particularly as @ashleyfrieze said he may have some.

src/main/java/com/greghaskins/spectrum/internal/Suite.java
```diff
@@ -255,16 +250,16 @@ private static void defaultChildRunner(final Suite suite, final RunNotifier runN
...
 
   @Override
   public boolean isEffectivelyIgnored() {
-    return ignored || !hasANonIgnoredChild();
+    return this.ignored || !hasANonIgnoredChild();
   }
 
   private boolean hasANonIgnoredChild() {
-    return children.stream()
+    return this.children.stream()
```
ashleyfrieze: Is the use of a redundant `this` really necessary?

Issue #100, opened 2017-03-02
The `ConfiguredBlock` and `BlockConfigurationChain` code violates the Open Closed Principle


ashleyfrieze:
We want to be able to add more pre-conditions and this would seem to require a change everywhere. Can the existing preconditions be refactored into something more object oriented?

Spoilers: YES

ashleyfrieze:
This was fixed by #101

PR #101, opened 2017-03-02
Modified `BlockConfigurationChain` to use typed objects 

ashleyfrieze:
to represent the configuration, so that it can be extended without (much) code change.

A proposed fix for #100

### DIFF
--- a/src/main/java/com/greghaskins/spectrum/BlockConfigurationChain.java
+++ b/src/main/java/com/greghaskins/spectrum/BlockConfigurationChain.java
@@ -1,37 +1,52 @@
 package com.greghaskins.spectrum;
 
+import com.greghaskins.spectrum.internal.BlockConfigurable;
 import com.greghaskins.spectrum.internal.BlockConfiguration;
 
+import java.util.stream.Stream;
+
+/**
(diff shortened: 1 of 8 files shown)


greghaskins:
@ashleyfrieze nice work! I like this approach much better than what we had. `with(tags("foo").and(focus()), () -> {})` reads much better, while also being open-closed. I'm basically ready to merge this PR, unless you have any further tweaks to make.

A couple observations:

1. It strikes me that we didn't have very strong coverage if this change doesn't require modifying any of the existing tests! 😯 At least we've got one in there now to test the chaining with `and()`.
2. One of my last remaining tasks for `1.1.0` is to make sure we don't expose any `internal` interfaces on the Public API (that way we can continue refactoring there with impunity after release). I've got it on my list to finish going through to make sure nothing `internal` is used as an input or output, and promote any necessary items to the main package. For this PR, I think that would only affect `BlockConfigurable`.

ashleyfrieze:
The actual chaining of multiple types of config had dropped out of tests. The reason the tests didn't change is that the facade via the Spectrum class wrapped it all up. In other words, the outward interface has barely changed and to testable behaviour is the same.

Merge if you like it!

PR #102, opened 2017-03-06
Continue API cleanup for 1.1.0

greghaskins:
This picks up where I left off with #88.

- [x] Pull Gherkin and Spec APIs into their own packages, with a single point of entry each. That will make it easier to `import static com.greghaskins.spectrum.something.Gherkin.*;` and get all the goodness.
    - [x] Gherkin interface
    - [x] Specification interface
    - [x] Common interface (e.g. tags, configuration, etc.) to shared location
- [x] Make sure nothing `internal` is involved in the public API (either as entry point, parameter, or return value). Some of these will need to move to main package as a result.
- [x] Resolve cyclic dependencies, at least involving the non-internal packages and their classes.
- [x] Include regression tests for `1.0.x` to guarantee backwards compatibility. This will also help resolve the coverage drop, since several methods are moving to new homes and the old homes are just pass-through.

Still work-in-progress. Open to feedback.


greghaskins:
Notable changes:

- The two primary entry-points are `com.greghaskins.spectrum.dsl.specification.Specification` and `com.greghaskins.spectrum.dsl.gherkin.Gherkin`. The old ones in `Spectrum.java` are still there for backward compatibility, but users should prefer the new, since it will make their `import` statements easier. No new public static functions should be added to Spectrum.class, they'll be added to the DSL classes moving forward.
- The filtering config now has an API similar to the block config: `filterRun(includeTags("foo").and(excludeTags("bar")))` instead of the previous ~`configure().includeTags("foo").excludeTags("bar")`~.
- Common functions for configuration are available from the `Configure` interface. This includes `with()`, `tags()`, `ignore()`, `focus()`, `filterRun()`, `includeTags()`, and `excludeTags()`. We may alias these into the DSL entry-point classes directly in the future (to make static imports even easier).
- Regression tests are being run against `1.0.0` as part of the build. Everything from the `1.0.0` suite is still passing (unmodified) with the new code.

TODO:

- We may need to take another pass at the docs to make sure they use the correct imports and API interfaces.

PR #111, opened 2017-04-18
Resolve the issues reported in #110 with bad reporting of errors

ashleyfrieze:
Fixes #110 

Errors were reported in the wrong order confusing IDE JUnit runners

Issue #110, opened 2017-03-23
Spectrum 1.1.0 displays test case nested structure wrong in Intellij Idea


anttiahonen:
Unfortunately version 1.1.0 seems to have introduced funky behavior in Intellij Idea test output structure in case of exception inside tests.

First the test code that throws exception:
[![spectrumkoodi.png](https://s28.postimg.org/y5mxz4dtp/spectrumkoodi.png)](https://postimg.org/image/d8qpugfsp/)
Second how Intellij Idea displays it with Spectrum 1.0.0
[![spectrum1.0.0.png](https://s30.postimg.org/b70h2uju9/spectrum1_0_0.png)](https://postimg.org/image/xiy9w8iy5/)
Third the same test with Spectrum 1.1.0
[![spectrum1.1.0.png](https://s2.postimg.org/xasg2w52h/spectrum1_1_0.png)](https://postimg.org/image/jh43duch1/)

Happens at least with Intellij Idea 2017.1 and 2016.3.5

ashleyfrieze:
This looks to be an issue with the way that Spectrum informs the `RunNotifier` about failure and where that happens in the process. The `RunNotifier` can be very picky about what you say to it and when. Looks like the more exotic `hook` mechanism we (and I mean _I_) added has some unexpected side effects when there are errors in the pre-test hooks.

@greghaskins do you mind if I investigate this over the next few days. Have any ideas of your own?

greghaskins:
@ashleyfrieze be my guest. Thanks for the help. I don't have any ideas off hand, but interested to see what you find.

ashleyfrieze:
I think it may also be related to #109 - I think we need to call `fireTestStarted` for a failing test to explain why it failed. My guess is that the solution to this may lay in abstracting away from `RunNotifier`, which can only help move toward test framework agnosticism.

I did a little spike, and it looks like it will need more than a one line fix :|

ashleyfrieze:
We're violating the contract with `RunNotifier` in a couple of ways. The hook mechanism would get very tangled if we didn't introduce an adapter.

I'm working on a branch here - https://github.com/ashleyfrieze/spectrum/tree/IntelliJErrorBug

Some of the problem is dealt with by 130b14873e91993a0c825505cc545af4406fabda

ashleyfrieze:
@anttiahonen if you would like to pull the branch behind #111 and try it on both of the issues you raised, I'd appreciate the feedback. I had reproduced your problem with this test:

[...]

Issue #109, opened 2017-03-23
Maven Surefire omits test method name from failed tests 


anttiahonen:
Running tests with Maven (Surefire) seems to leave out the class name, describe and it-blocks from the test summary. Here is an example:

```
Results :

Failed tests: 

//This is a failing spectrum test, only assertion showing
  expected:<24000> but was:<2400>

//This is a regular failing JUnit 4 test
PricingServiceTest.testMultiSectionApplication:89 expected:<22500> but was:<220500>

Tests run: 160, Failures: 2, Errors: 0, Skipped: 2
```

As one of the best aspects of spec-style testing is the readable test output, this is a little inconvenient.

ashleyfrieze:
I recommend closing this because I think that #112 fixed it.

greghaskins:
Closing this. Looks like Spectrum is doing better with Surefire in the current release.

Example:
```java
@RunWith(Enclosed.class)
public class Tests {

    @RunWith(JUnit4.class)
    public static class TraditionalTests {
        @Test
        public void foo() throws Exception {
            assertEquals("an expected value", "an incorrect value");
        }
    }

    @RunWith(Spectrum.class)
    public static class SpectrumTests {

        {
            describe("The thing", () -> {

                it("does something", () -> {
                    assertEquals("someExpectedValue", "someActualValue");
                });

            });
        }

    }
}
```

Maven console output:

```
Results :

Failed tests:   does something(The thing): expected:<some[Expected]Value> but was:<some[Actual]Value>
  foo(com.example.Tests$TraditionalTests): expected:<an [expected] value> but was:<an [incorrect] value>

Tests run: 2, Failures: 2, Errors: 0, Skipped: 0
```

Surefire definitely makes some formatting decisions based on traditional JUnit test naming, but at least all the context comes through in one place or another.

PR #112, opened 2017-04-23
Fixes for `beforeAll` failures

ashleyfrieze:
Fixes for #110 and #109 further to #111 not quite fixing `beforeAll`.

Added additional tests and resolved a couple of minor glitches in docs and code for the JUnit rule support.


ashleyfrieze:
Squashed the commit to make for a nicer merge... no pressure @greghaskins  ;)

PR #118, opened 2017-08-24
Removed thread-segregation from `Variable` and `let`.

### LINKED ISSUE #115: Variables+awaitility only work if used in the same thread
i am not sure what you can do on your end, but thought I would report it anyhow.   the following test case fails because Variables use a ThreadLocal, and awaitility by default happens to run in a separate one.  The only way it works is if the user does an 
`await().pollInsSameThread()`.

```
java.lang.AssertionError: 
Expected: not (null or an empty string)
     but: was null
Expected :not (null or an empty string)
     
Actual   :null
```

```
package test;

import com.greghaskins.spectrum.Spectrum;
import com.greghaskins.spectrum.Variable;
import org.junit.runner.RunWith;

import static com.greghaskins.spectrum.Spectrum.describe;
import static com.greghaskins.spectrum.Spectrum.it;
import static org.awaitility.Awaitility.await;
import static org.hamcrest.MatcherAssert.assertThat;
import static org.hamcrest.Matchers.isEmptyOrNullString;
import static org.hamcrest.Matchers.not;

@RunWith(Spectrum.class)
public class VariableTest {{
  describe("suite", () -> {
    final Variable<String> var = new Variable<>();
    it("test", () -> {
      var.set("value");
      await().until(() -> {
        assertThat(var.get(), not(isEmptyOrNullString()));
        return true;
      });
    });
  });
}}
```

ashleyfrieze:
Fixes #115 and supersedes #117. This removes (for now) the artificial thread-separate introduced with the `Variable` class and `let hook` added in release `1.1.0`.

I will work out how to make parallel execution thread safe in its own PR.

### DIFF
--- a/src/main/java/com/greghaskins/spectrum/Variable.java
+++ b/src/main/java/com/greghaskins/spectrum/Variable.java
@@ -10,7 +10,7 @@ import java.util.function.Supplier;
  */
 public final class Variable<T> implements Supplier<T> {
 
-  private ThreadLocal<T> value = new ThreadLocal<>();
+  private T value;
 
   /**
    * Create a Variable with a {@code null} initial value.
@@ -33,7 +33,7 @@ public final class Variable<T> implements Supplier<T> {
    */
   @Override
   public T get() {
-    return this.value.get();
+    return this.value;
   }
 
   /**
@@ -42,6 +42,6 @@ public final class Variable<T> implements Supplier<T> {
    * @param value new value
    */
   public void set(final T value) {
-    this.value.set(value);
+    this.value = value;
   }
 }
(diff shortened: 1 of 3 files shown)


greghaskins:
@ashleyfrieze looks great. +1 for the tests too. I'll merge and push a new version by tomorrow.

ashleyfrieze:
I think you know me well enough to know I'm a test first guy :)

What kind of test framework would not have tests for discovered bugs in boundary conditions!!!?

This is going to make parallel tests much harder to implement.

ashleyfrieze:
By the way, coverage is down because there is less code rather than less coverage.

Issue #74, opened 2016-12-02
Add `timeout` pre-condition


ashleyfrieze:
Similar to JUnit's `@Test(timeout=1234)` this is a suggestion to extend the current pre-condition functionality a bit like this:

```java
it("will run but not for long", with(timeout(1234), ()->{
   // spec code
}));
```

This is quite easy to do, since the `PreConditionBlock` can just handle it for us, wrapping the real block in a timing out sentinel. But there's a wrinkle. What does:

```java
describe("this has a timeout so what does it mean?", with(timeout(1234), ()-> {
   // specs and describes go in here
   it("tests", () -> {});
}));
```

If we think it means that the whole parent block runs within a timeout, then the behaviour would happen for free. I think that's the least valuable and least expected answer. More likely it means that the above is the equivalent of:

```java
// any time out put in a describe is sent down to its next most atomic children and not executed at parent level
describe("this has a timeout so what does it mean?", ()-> {
   // specs and describes go in here
   it("tests",  with(timeout(1234, () -> {}));
}));
```

Thoughts on:

* Is this valuable as a feature (clue: yes :) )
* How do we think it should pass timeouts around the tree?
* Is the above syntax and re-use of the `PreConditions` a good approach?

richdouglasevans:
> Is this valuable as a feature?

Yes.

> How do we think it should pass timeouts around the tree?

Dunno; having a play now.

> Is the above syntax and re-use of the PreConditions a good approach?

Sounds good.

> If we think it means that the whole parent block runs within a timeout...

That would be surprising. Your example where such preconditions are defaults that are applied to atomic children makes intuitive sense to me.

ashleyfrieze:
Thanks Rich. If you fancy a play look at the hooks mechanism. This will work well for _describe_ but poorly for _it_.

I was wondering whether a hook for single _it_s might be in order.

ashleyfrieze:
Looking at this simply, if `timeout` were a precondition - i.e. a part of `BlockConfiguration` that was metadata like `tags` is, then we could create a simple hook which always applied to all children in spectrum and, when there was a timeout, ran the particular child in a thread with a sentinel.

So, expand `BlockConfiguration` and make a `timeoutHook`. However...

Oh boy does this violate the open-closed principle! Because if we came up with another, similar, requirement, we'd have to make two things again. Every time.

[...]

PR #119, opened 2017-08-31
Spectrum to allow spec timeout to be specified throughout the hierarchy

ashleyfrieze:
An implementation to address #74 

Still WIP at the moment, feedback appreciated.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -1,0 +1,62 @@
+## Configuration
+
+Metadata to influence the execution can be added anywhere in the test hierarchy. Wherever you can add a block containing suites, tests, or a test body, you can also tag that block with configuration to apply from that point in the hierarchy downwards.
+
+The common use case for this would be selective running, where tagging with `ignore` or `tags` would control which parts of the hierarchy are executed in a given test run. See [selective running](FocusingAndIgnoring.md) for more information.
+
+### Adding configuration to blocks
+
+The basic format is to take the block inside a `describe` or `it` block (and others) and wrap it inside a `with`. E.g.
+
+```java
+describe("Some parent", () -> {
+  it("tests something", () -> {});
+});
+
+// might be tagged at parent level and become
+describe("Some parent", with(focus(), () -> {
+  it("tests something", () -> {});
+}));
+```
+
+Configurations can be chained by using the `and` function, for example this spec has both tags and a timeout:
+
+```java
+describe("A suite", with(tags("someTag").and(timeout(10, TimeUnit.SECONDS)), () -> {
+  it("will test something in time", () -> {
+    ...
+  });
+}));
+```
+
+`and` can be chained further:
+
+```java
+describe("A suite", with(
+  tags("someTag")
+    .and(timeout(ofSeconds(10)))
+    .and(focus()),
+    () -> {
+      ...
+    }
+));
+```
+
+### Scope of configuration
+
+A configuration applies to the current node in the hierarchy and its children. With `ignore`, the effect of ignoring a parent is to ignore all children. With `tag`, the tagging is cumulative. With `timeout` the timeout settings in a parent propagate down to all descendants but can be superseded by a child's own configuration.
+
+The general rule is that the configuration applies to the whole hierarchy and can only be added to or superseded by children that are allowed to execute.
+
+### Configurations available
+
+In addition to:
+
+- `ignore()` - ignore this test
+- `ignore(String reason)` - ignore this test with a reason
+- `focus()` - run focused specs only, especially this
+- `tags(String ... tags)` - tag the spec with labels
+
+there is also:
+
+- `timeout(Duration timeout)` - make the test fail if it takes too long - see [Timeout](Timeout.md)
--- a/docs/FocusingAndIgnoring.md
+++ b/docs/FocusingAndIgnoring.md
@@ -100,8 +100,18 @@ describe("Ignored specs", () -> {
 ```
 #### Ignoring Specs Spectrum style
 
-Spectrum allows you to specify preconditions on a block. These preconditions can include tagging.
-The `with(BlockConfigurationChain,Block)` function is used to annotate a block with preconditions. Preconditions include:
+Spectrum allows you to specify preconditions for execution as part of the configuration of a block. These can include tagging.
+The `with(BlockConfigurationChain,Block)` function is used to annotate a block, for example:
+
+```java
+describe("an ignored suite", with(ignore(), () -> {
+  ...
+}));
+```
+
+For the full capabilities of adding configuration to blocks see [Configuration](Configuration.md).
+
+Available tagging for ignoring:
 
 - `ignore()` - ignore this test
 - `ignore(String reason)` - ignore this test and provide a reason to the reader
--- a/docs/README.md
+++ b/docs/README.md
@@ -15,7 +15,7 @@ Spectrum also supports:
 - Rigorous error handling and reporting when something unexpected goes wrong
 - Compatibility with most existing JUnit tools; no configuration required
 - Plugging in familiar JUnit-friendly libraries like `MockitoJUnit` or `SpringJUnit` [via JUnit `@Rule`s handling](JunitRules.md).
-- Tagging specs for [selective running](FocusingAndIgnoring.md)
+- Tagging specs for [selective running](FocusingAndIgnoring.md) or adding [configuration](Configuration.md) including [timeouts](Timeout.md)
 - Mixing Spectrum tests and normal JUnit tests in the same project suite
 - RSpec-style `aroundEach` and `aroundAll` hooks for advanced users and plugin authors
 
@@ -25,4 +25,4 @@ Unlike some BDD-style frameworks, Spectrum is _only_ a test runner. Assertions, 
 
 ## Getting Started
 
-See the [quickstart waklthrough](QuickstartWalkthrough.md) and the docs for writing [Specification-style](SpecificationDSL.md) or [Gherkin-style](GherkinDSL.md) tests.
+See the [quickstart walkthrough](QuickstartWalkthrough.md) and the docs for writing [Specification-style](SpecificationDSL.md) or [Gherkin-style](GherkinDSL.md) tests.
--- a/src/main/java/com/greghaskins/spectrum/Configure.java
+++ b/src/main/java/com/greghaskins/spectrum/Configure.java
@@ -4,11 +4,14 @@ import com.greghaskins.spectrum.internal.DeclarationState;
 import com.greghaskins.spectrum.internal.configuration.BlockFocused;
 import com.greghaskins.spectrum.internal.configuration.BlockIgnore;
 import com.greghaskins.spectrum.internal.configuration.BlockTagging;
+import com.greghaskins.spectrum.internal.configuration.BlockTimeout;
 import com.greghaskins.spectrum.internal.configuration.ConfiguredBlock;
 import com.greghaskins.spectrum.internal.configuration.ExcludeTags;
 import com.greghaskins.spectrum.internal.configuration.IncludeTags;
 import com.greghaskins.spectrum.internal.junit.Rules;
 
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 
 public interface Configure {
@@ -18,8 +21,8 @@ public interface Configure {
 
   /**
    * Surround a {@link Block} with the {@code with} statement to add
-   * preconditions and metadata to it. E.g. <code>with(tags("foo"), () -&gt; {})</code>.<br>
-   * Note: preconditions and metadata can be chained using the
+   * configuration and metadata to it. E.g. <code>with(tags("foo"), () -&gt; {})</code>.<br>
+   * Note: configuration metadata can be chained using the
    * {@link BlockConfigurationChain#and(BlockConfigurationChain)} method. E.g.
    * <code>with(tags("foo").and(ignore()), () -&gt; {})</code>
    *
@@ -30,6 +33,7 @@ public interface Configure {
    * @see #ignore()
    * @see #focus()
    * @see #tags(String...)
+   * @see #timeout(Duration)
    */
   static Block with(final BlockConfigurationChain configuration, final Block block) {
     return ConfiguredBlock.with(configuration.getBlockConfiguration(), block);
@@ -93,6 +97,16 @@ public interface Configure {
    */
   static BlockConfigurationChain focus() {
     return new BlockConfigurationChain().with(new BlockFocused());
+  }
+
+  /**
+   * Apply timeout to all leaf nodes from this level down. Can be superseded by a lower level having its
+   * own timeout.
+   * @param timeout the amount of the timeout
+   * @return a chainable configuration that will apply a timeout to all leaf nodes below
+   */
+  static BlockConfigurationChain timeout(Duration timeout) {
+    return new BlockConfigurationChain().with(new BlockTimeout(timeout));
   }
 
   /**
--- a/src/main/java/com/greghaskins/spectrum/internal/Child.java
+++ b/src/main/java/com/greghaskins/spectrum/internal/Child.java
@@ -3,6 +3,7 @@ package com.greghaskins.spectrum.internal;
 import com.greghaskins.spectrum.Block;
 import com.greghaskins.spectrum.internal.configuration.ConfiguredBlock;
 import com.greghaskins.spectrum.internal.configuration.TaggingFilterCriteria;
+import com.greghaskins.spectrum.internal.hooks.Hook;
 
 import org.junit.runner.Description;
 import org.junit.runner.notification.Failure;
@@ -39,20 +40,5 @@ public interface Child {
    */
   default boolean isLeaf() {
     return false;
-  }
-
-  /**
-   * Gets the object to be filtered appropriately with its preconditions.
-   * @param block the block that will be executed by the child - this may be of
-   *              type {@link ConfiguredBlock}.
-   * @param taggingFilterCriteria the tagging state in the parent of this suite or spec.
-   *                     This is used to determine what filters apply to the block
-   * @return the child itself for fluent calling
-   */
-  default Child applyPreconditions(final Block block,
-      final TaggingFilterCriteria taggingFilterCriteria) {
-    ConfiguredBlock.findApplicablePreconditions(block).applyTo(this, taggingFilterCriteria);
-
-    return this;
   }
 }
--- a/src/main/java/com/greghaskins/spectrum/internal/LeafChild.java
+++ b/src/main/java/com/greghaskins/spectrum/internal/LeafChild.java
@@ -1,0 +1,16 @@
+package com.greghaskins.spectrum.internal;
+
+import com.greghaskins.spectrum.internal.hooks.HookContext;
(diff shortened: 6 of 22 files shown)


docs/Configuration.md
```diff
@@ -0,0 +1,63 @@
+## Configuration
```
ashleyfrieze: This wasn't incredibly well documented.

docs/FocusingAndIgnoring.md
```diff
@@ -100,8 +100,18 @@ describe("Ignored specs", () -> {
 ```
 #### Ignoring Specs Spectrum style
 
-Spectrum allows you to specify preconditions on a block. These preconditions can include tagging.
-The `with(BlockConfigurationChain,Block)` function is used to annotate a block with preconditions. Preconditions include:
+Spectrum allows you to specify preconditions for execution as part of the configuration of a block. These can include tagging.
```
ashleyfrieze: Removing `precondition` as a keyword in relation to the block metadata, which we're really calling `BlockConfiguration`

docs/README.md
```diff
@@ -25,4 +25,4 @@ Unlike some BDD-style frameworks, Spectrum is _only_ a test runner. Assertions,
 
 ## Getting Started
 
-See the [quickstart waklthrough](QuickstartWalkthrough.md) and the docs for writing [Specification-style](SpecificationDSL.md) or [Gherkin-style](GherkinDSL.md) tests.
+See the [quickstart walkthrough](QuickstartWalkthrough.md) and the docs for writing [Specification-style](SpecificationDSL.md) or [Gherkin-style](GherkinDSL.md) tests.
```
ashleyfrieze: A _tyop_

src/main/java/com/greghaskins/spectrum/Configure.java
```diff
@@ -96,6 +100,27 @@ static BlockConfigurationChain focus() {
...
+   * own timeout.
+   * @param timeout the amount of the timeout
+   * @param timeUnit the unit of the timeout
+   * @return a chainable configuration that will apply a timeout to all leaf nodes below
+   */
+  static BlockConfigurationChain timeout(long timeout, TimeUnit timeUnit) {
+    return new BlockConfigurationChain().with(new BlockTimeout(timeout, timeUnit));
+  }
+
+  /**
```
ashleyfrieze: Timeouts. We waited long enough for them.

src/main/java/com/greghaskins/spectrum/internal/Child.java
```diff
@@ -40,19 +41,4 @@ default boolean isAtomic() {
...
-   * @param taggingFilterCriteria the tagging state in the parent of this suite or spec.
-   *                     This is used to determine what filters apply to the block
-   * @return the child itself for fluent calling
-   */
-  default Child applyPreconditions(final Block block,
-      final TaggingFilterCriteria taggingFilterCriteria) {
-    ConfiguredBlock.findApplicablePreconditions(block).applyTo(this, taggingFilterCriteria);
-
-    return this;
-  }
```
ashleyfrieze: This didn't have to be here. Its implementation has moved.

src/main/java/com/greghaskins/spectrum/internal/LeafChild.java
```diff
@@ -0,0 +1,14 @@
+package com.greghaskins.spectrum.internal;
+
+import com.greghaskins.spectrum.internal.hooks.Hook;
+
+/**
+ * Interface for a child that is also a leaf node in the test hierarchy.
+ */
+public interface LeafChild extends Child {
```
ashleyfrieze: This _leaf_ interface allows us to use `instanceof` and a cast to specific type in order to work with a leaf node to apply leaf-level hooks. If we didn't do that we'd end up adding an unreachable method to `Child` for some of its descendants - the code coverage led me to this approach.

src/main/java/com/greghaskins/spectrum/internal/Spec.java
```diff
@@ -1,18 +1,23 @@
...
+import com.greghaskins.spectrum.internal.hooks.Hooks;
 
 import org.junit.runner.Description;
 import org.junit.runner.notification.Failure;
 
-final class Spec implements Child {
+final class Spec implements LeafChild {
 
-  private final NotifyingBlock block;
+  private final Block block;
```
ashleyfrieze: The block is kept here in its raw form, as the leaf hooks will wrap it with reporting.

src/main/java/com/greghaskins/spectrum/internal/Spec.java
```diff
@@ -31,8 +36,8 @@ public void run(final RunReporting<Description, Failure> notifier) {
       return;
     }
 
-    this.block.run(this.description, notifier);
-
+    // apply leaf hooks around the inner block
+    leafHooks.runAround(this.description, notifier, block);
```
ashleyfrieze: Leaf hooks now always run, even when there are none.

src/main/java/com/greghaskins/spectrum/internal/Suite.java
```diff
@@ -30,7 +31,7 @@
   private boolean ignored;
 
   private final TaggingFilterCriteria tagging;
-  private BlockConfiguration preconditions = BlockConfiguration.defaultConfiguration();
+  private BlockConfiguration configuration = BlockConfiguration.defaultConfiguration();
```
ashleyfrieze: Reads better after rename

src/main/java/com/greghaskins/spectrum/internal/Suite.java
```diff
@@ -70,9 +71,10 @@ public Suite addSuite(final String name) {
   }
 
   private Suite addSuite(final String name, final ChildRunner childRunner) {
-    final Suite suite =
-        new Suite(Description.createSuiteDescription(sanitise(name)), this, childRunner,
-            this.tagging.clone());
+    final Suite suite = new Suite(Description.createSuiteDescription(sanitise(name)), this, childRunner,
+        this.tagging.clone());
+
+    suite.inheritConfigurationFromParent(configuration.forChild());
```
ashleyfrieze: We were not propagating configurations especially cleanly. It's now worked in.

src/main/java/com/greghaskins/spectrum/internal/configuration/BlockConfiguration.java
```diff
@@ -12,12 +12,12 @@
...
   /**
-   * Combine provided precondition objects together.
+   * Combine provided configuration objects together.
    *
    * @param conditions to combine
-   * @return a combination of all preconditions as a new object
+   * @return a combination of all configurations as a new object
    */
-  static BlockConfiguration merge(BlockConfiguration... conditions) {
+  public static BlockConfiguration merge(BlockConfiguration... conditions) {
```
ashleyfrieze: Opening this up seemed to make sense.

src/main/java/com/greghaskins/spectrum/internal/configuration/BlockTimeout.java
```diff
@@ -0,0 +1,76 @@
...
+    // my timeout supersedes any inherited timeout
+
+    return this;
+  }
+
+  private Hook timeoutHook() {
+    return withUnreportedInnerBlock((description, reporting, block) -> {
+      FailOnTimeout.builder().withTimeout(timeout, timeUnit)
+          .build(statementOf(block))
+          .evaluate();
```
ashleyfrieze: Essentially this is `JUnit`'s own implementation that I'm using here... it caused a few glitches, hence some reworking of the way that `NotifyingBlock` is used inside tests, since the multi-threaded nature of the above mechanism gave me some `InterruptedException` headaches.

src/main/java/com/greghaskins/spectrum/internal/configuration/TaggingFilterCriteria.java
```diff
@@ -63,15 +63,10 @@ private boolean compliesWithRequired(Collection<String> tags) {
...
 
-  private String[] fromSystemProperty(final String property) {
+  private Stream<String> fromSystemProperty(final String property) {
     return Optional.ofNullable(System.getProperty(property))
-        .map(string -> string.split(TaggingFilterCriteria.TAGS_SEPARATOR))
-        .filter(TaggingFilterCriteria::notArrayWithEmptyValue)
-        .orElse(new String[0]);
+        .map(string -> Arrays.stream(string.split(TaggingFilterCriteria.TAGS_SEPARATOR))
+            .filter(tag -> !tag.isEmpty()))
+        .orElse(Stream.empty());
```
ashleyfrieze: Simplified in an attempt to improve code coverage.

src/main/java/com/greghaskins/spectrum/internal/hooks/Hook.java
```diff
@@ -27,11 +27,40 @@ void accept(final Description description, final RunReporting<Description, Failu
       final Block block) throws Throwable;
 
   /**
+   * Override to return true if the inner block cannot report its own errors for some reason.
+   * @return true to suppress wrapping inner block in self-reporting
+   */
+  default boolean requiresUnreportedInnerBlock() {
+    return false;
+  }
```
ashleyfrieze: Hooks were being heavily wrapped in exception handling/reporting, but this is not necessary for all hooks.

src/main/java/com/greghaskins/spectrum/internal/hooks/Hooks.java
```diff
@@ -95,25 +96,34 @@ private Hook createChainOfResponsibility(Variable<Boolean> hooksRememberedToRunT
...
-        wrapWithReporting(description, reporting,
+        conditionallyWrapWithReporting(inner, description, reporting,
             () -> inner.accept(description, reporting, block)));
   }
 
+  private static Block conditionallyWrapWithReporting(final Hook forHook, final Description description,
+      final RunReporting<Description, Failure> reporting, final Block innerBlock) {
+    if (forHook.requiresUnreportedInnerBlock()) {
+      return innerBlock;
+    }
```
ashleyfrieze: This is kind of a workaround and kind of an efficiency - some hooks can ask for no wrapping with additional notification/error handling.

src/test/java/specs/JUnitRuleExample.java
```diff
@@ -27,6 +27,16 @@
...
+  // alternative morphology of providing a rule - see http://junit.org/junit4/javadoc/4.12/org/junit/Rule.html
+  public static class TempFolderRuleProvidedViaMethodMixin {
+    private TemporaryFolder tempFolderRule = new TemporaryFolder();
+
+    @Rule
+    public TemporaryFolder getFolder() {
+      return tempFolderRule;
+    }
+  }
+
```
ashleyfrieze: This permutation of JUnit `@Rule` wasn't tested before, though supported in the code.

src/test/java/specs/TimeoutSpecs.java
```diff
@@ -0,0 +1,92 @@
...
+import org.junit.runner.RunWith;
+
+import java.util.concurrent.TimeUnit;
+
+@RunWith(Spectrum.class)
+public class TimeoutSpecs {
+  {
+    describe("A suite with timeouts", () -> {
+      it("will allow things to pass if they run quicker than the timeout", () -> {
+        final Result result = SpectrumHelper.run(() -> {
```
ashleyfrieze: Using this approach rather than method-level classes, as I think it reads better... feedback appreciated.

ashleyfrieze:
@greghaskins - I'm hoping you'll find this an easy PR to approve, given I'm just following the `with`/`and` stuff we set up for `ignore`/`focus` and `tags`.

ashleyfrieze:
One TODO left - I want to make the `Hooks` at _leaf_ level unable to do weird stuff with `timeout`. Need to decide how to do that. Probably limit leaf hooks from requiring error notification/reporting.

src/test/java/specs/TimeoutSpecs.java
greghaskins: Agreed. I added that test helper for this exact purpose, as I got sick of writing nested classes.

src/test/java/specs/TimeoutSpecs.java
```diff
@@ -0,0 +1,111 @@
...
+          describe("Suite with low timeout", with(timeout(1, TimeUnit.MILLISECONDS), () -> {
+            describe("with child suite", () -> {
+              it("has a more generous timeout", with(timeout(1, TimeUnit.MINUTES), () -> {
+                Thread.sleep(100);
+              }));
+            });
+          }));
+        });
+        assertThat(result.getFailureCount(), is(0));
```
greghaskins: I noticed all the assertions here are confirming that tests fail/pass in the right scenarios. What about the messaging and user feedback that goes along with that? I haven't reviewed the implementation yet, but I assume there are some helpful messages that show up when one of these fail ("Timeout expired" or something to that effect).

greghaskins:
I very much like this feature. 👍 Especially for being able to use a specified time unit.

Question: why not use the [Java 8 Duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html) object? `Duration.ofSeconds(5)` is pretty nice, which could be further simplified like `with(timeout(ofSeconds(5))) ...` after importing things statically.

I, for one, always lose time trying to count zeros and mentally convert between time units in my tests, so I almost feel like we should make units mandatory for readability. (Using a`Duration` or just making that `TimeUnit` parameter required would accomplish that.)

src/test/java/specs/TimeoutSpecs.java
greghaskins: Ah, I see it's leveraging JUnit4's [`FailOnTimeout` helper](https://github.com/junit-team/junit4/blob/master/src/main/java/org/junit/internal/runners/statements/FailOnTimeout.java). Might still be worth verifying some of that behavior in case we end up switching to something else down the line (JUnit5 might have a different helper, or we write our own).

ashleyfrieze: We could wrap the helper's exception with a Spectrum one and then verify that, giving a seamless switch between implementations, or we could not promise anything special beyond _it fails_. What do you think?

ashleyfrieze:
For full polyglot-like _it is how you are expecting_ for JUnit aficionados, the pure millisecond version is most natural, followed by the two parameter version as used in the timeout rule. I really like your Duration version, though. Could I just add it as an overload?

ashleyfrieze:
Note: there will be a squash and rebase. Other than that, I'm pretty much ready with this, beyond feedback driven changes.

ashleyfrieze:
@greghaskins - here is why I think we should NOT implement `Duration` in `timeout`, even though I think it's a good idea in theory.

1. The low-level implementation has to be turned into either _milliseconds_ or `long` + `TimeUnit` to be compatible with the Java thread waiting mechanism
2. `Duration` is weirdly in _seconds_ and _nanoseconds of a second_ which creates some awkward calculation
3. To test the calculation behaviourally with actual timeouts would suck so...
4. We could make timeout setting use interfaces JUST because of `Duration` and test with mocks, but that would suck
5. We could make the timeout conversion from `Duration` convert to an object with a `long` and a `TimeUnit` in it, so we can unit test the POJO... and... 
6. We'd have just created a custom `Duration` class of our own that we're not even going to let people use!

It seems a bit futile. In general, Java waits are expressed in milliseconds, or in `long` + `TimeUnit`. See `System.currentTimeMillis` and `Object#wait` and `Thread.sleep`.

src/main/java/com/greghaskins/spectrum/internal/junit/TimeoutWrapper.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.junit.internal.runners.statements.FailOnTimeout;
+
+import java.time.Duration;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * Wrap JUnit's timeout mechanism as a {@link NonReportingHook}.
+ */
+public interface TimeoutWrapper {
```
ashleyfrieze: It kind of made sense to have the `BlockTimeout` not do its own implementation of timeout and to move our dependency on JUnit4 into the `junit` package. This means we could implement the timeout hook another way in future and only change the factory method used... probably a sensible bit of design.

src/main/java/com/greghaskins/spectrum/internal/junit/TimeoutWrapper.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  /**
+   * Apply a timeout expressed as a duration to a builder of a {@link FailOnTimeout} object.
+   * @param builder to modify
+   * @param timeout duration of the timeout
+   * @return the builder input - for fluent use.
+   */
+  static FailOnTimeout.Builder withAppliedTimeout(FailOnTimeout.Builder builder, Duration timeout) {
+    builder.withTimeout(timeout.toNanos(), TimeUnit.NANOSECONDS);
```
ashleyfrieze: While I had reservations about do this this way, it works and is simple. It _could_ be made more clever with different sorts of time range, and this function is easily tested, but it's not necessary.

src/test/java/specs/JUnitRuleExample.java
```diff
@@ -77,6 +87,15 @@ public static void beforeClass() {
...
     });
+
+    describe("A spec with a rule mix-in where the rule is provided by method", () -> {
+      Supplier<TempFolderRuleProvidedViaMethodMixin> tempFolderRuleMixin =
+          junitMixin(TempFolderRuleProvidedViaMethodMixin.class);
+
+      it("has access to an initialised object", () -> {
+        assertNotNull(tempFolderRuleMixin.get().getFolder().getRoot());
+      });
+    });
```
ashleyfrieze: Another missing bit of test

src/test/java/specs/TimeoutSpecs.java
ashleyfrieze: The nested classes were a rather surprising bit of java syntax - cleverly found!

src/test/java/specs/TimeoutSpecs.java
ashleyfrieze: I'm inclined not to worry about the exact failure - only that it fails.

ashleyfrieze:
> @greghaskins - here is why I think we should NOT implement Duration in timeout, even though I think it's a good idea in theory.

Ok, so it turned out that, after I gave it some more thought, I changed my mind. This was because:

1. I want @greghaskins to have every reason to accept this PR quickly so I can do other stuff
2. I trusted an easy way to convert from `Duration` to `long`+`TimeUnit` via converting everything to nanoseconds
3. I thought the separation of concerns between Spectrum with `Duration` and `JUnit`'s internals in `long`+`TimeUnit` was good for ~me~ us.

I think the result is better.

greghaskins:
Looks good. Merging.

A few edge cases come to mind that I didn't see tests for and haven't tried yet myself; curious how they would be handled:

1. What if the slowdown was in a `beforeEach` /`afterEach` block? I would expect this looks the same as if the test body itself was slow.
2. Where does the time for `beforeAll`/ `afterAll` count? For example, if my timeout is 5 seconds and `beforeAll` takes 7 seconds, what happens? Then, what about multiple and nested `beforeAll`/ `afterAll` blocks?
4. Do `aroundEach` and `aroundAll` work with this? I'm guessing yes.
3. Does the `timeout` hook interrupt long-running tests, or just fail if they go too long?

ashleyfrieze:
@greghaskins - let's continue this discussion under #74

PR #120, opened 2017-09-01
Release version 1.2.0

greghaskins:


ashleyfrieze:
:+1: from me.

Makes me think we should have a `.sh` script for updating this in all the necessary places. Looks a bit of a nuisance to do!

greghaskins:
It's not too bad, but full automation is always the dream. I have a [wiki page](https://github.com/greghaskins/spectrum/wiki/Publishing-a-release) for reminding myself and keeping it consistent, at least.

Issue #103, opened 2017-03-13
Provide proxy interface as well as `Supplier` for `let` and `junitMixin`


ashleyfrieze:
Where it's possible, it would be a brilliant bit of syntactic sugar to be able to replace

```java
Supplier<MyThing> myThing = let(() -> new MyThing());

// in test code
Foo foo = myThing.get().getMyProperty();
```

with something where there's less indirection - say:

```java
MyThingInterface myThing = let(MyThingInterface.class, () -> new MyThing());

Foo foo = myThing.getMyProperty();
```

where `MyThingInterface` is an interface implemented by `MyThing`. This would probably be even more useful in the `junitMixin` where you may be likely to be reading properties of properties and the pattern would encourage an interface facade against all the commonly accessed things in the mixin, which can then easily be used.

@greghaskins - what do you think of this? I think it can be done in about 20 lines of neat code and added as an overload to what we have - essentially it's a shim to convert a `Supplier<T>` to an `R` where `T extends R` and the object of `R` is just a dynamic proxy to the object behind `Supplier.get`.

ashleyfrieze:
@greghaskins do you even slightly care about this? I'm not sure I do anymore :)

greghaskins:
Let's put this idea on the shelf until we get lots of complaints about the syntax. Some boilerplate is pretty much expected in the Java world these days, so I don't see the extra `.get()` and `.set()` as too painful. If people want really slick syntax, they would probably already be using Groovy/Scala/Kotlin, etc.

ashleyfrieze:
Interestingly one of the comments from that article I fwded was on the clunkiness of this... might be worth keeping open at low priority.

ashleyfrieze:
@greghaskins - is this still closed?

basdijkstra:
Again, since @ashleyfrieze invited me to comment, here's my $0.02:

I'd like to see this implemented anyway. Don't think that 'if people expect clean code they'd move to Groovy/Scala/Kotlin' is a good reason not to strive for clean code in Java :)

Plus, it would make using Spectrum that much easier to explain to people that do not necessarily have a ton of programming experience.

greghaskins:
Alright, I'm on board. Let's continue the discussion over in PR #122.

PR #122, opened 2017-09-04
Experimental feature for unboxing.

ashleyfrieze:
This is what I had in mind for #103.

@greghaskins - does it majorly suck, or does it add value?


greghaskins:
Nice.

Does it work with concrete types or only interfaces? I'm thinking people would really like to use this for mocks, which are frequently (with Mockito at least) concrete types.

Still mulling over the `unbox()` entry point. Not the worst thing, but another import/function for people to learn. My first thought was overloading `let` like `let(Something.class, () -> new Something())`. 

There may also be some sneaky tricks to even avoid passing in that `Class` object; I feel like maybe Hamcrest or something has some code that just works with generics somehow – `unbox(let(() -> new Something());` would be pretty slick.

ashleyfrieze:
The inbox function as provided can be used to surround any supplier, of which both let and junitMixin are such... but we could provide overloads to wrap it further.

It can ONLY work for interfaces. This is a technical constraint to do with dynamic proxies, but also makes sense if you think about how concrete types can have accessible instance variables and multiple constructors, which even Mockito has limitations over.

The type must be specified in the call because the stuff you mentioned about inferring type only works for generic types themselves, not for actual Class objects. In other words, to know at run time the interface to supply, you need a Class object. Even Mockito.mock requires a class object as input.

ashleyfrieze:
> My first thought was overloading let like let(Something.class, () -> new Something()).

I've extended my POC to have the equivalent of this - I reverse the order of parameters from the above, because that's what felt natural, but it's easily refactored. I did it for `junitMixin` and `let`, but left the `unboxer` documented and public as this tool could conceivably be used by end-users for their own combinations of things.

How does it look?

greghaskins:
Understood ✅ . A couple thoughts:

- If it only works with interfaces, we should fail fast (throw an `IllegalArgumentException`) to help people realize that quickly, rather than getting a stacktrace from deep in the standard library.
- Since this is experimental, and limited to just interfaces (for now – back of my mind is already wondering how to use ByteBuddy down the road...), let's just use a single entry point like `unbox`. If this feature takes off and people really like it, we can add the convenience overloads of `let` and `junitMixin`.
- I don't know what the best name for `unbox()` would be, but I can't think of anything significantly better. `unwrap()`, `proxy()`, `get()`, and `from()` come to mind; I'm not in love with any of them.

ashleyfrieze:
So, I should remove the overloads of let and junitmixin? It is easy to check if the passed in class is an interface.

Naming the method was hard. It's an auto get or an unwrapper or a de supplier or... nothing sounded good. Accessor?

Issue #121, opened 2017-09-01
Provide native support for expected exceptions


ashleyfrieze:
There are a few ways to test for some code ending in an exception.

- Use a framework like `AssertJ` with `assertThatThrownBy`

E.g.
```java
it("expects a boom", () -> {
   assertThatThrownBy(() -> { throw new Exception("boom!"); }).isInstanceOf(Exception.class)
                                                             .hasMessageContaining("boom");
}
```

- Employ the `ExpectedException` rule from JUnit either with `junitMixin` or even raw with `@Rule`

E.g.

```java
@Rule
public ExpectedException expectedException = ExpectedException.none();

{
   it("expects a boom", () -> {
         expectedException .expect(NullPointerException.class);
         throw new NullPointerException();
   });
}
```

- You could even do `aroundEach`

```java
describe("These all need an exception to be thrown", () -> {
   aroundEach(block -> {
      try {
         block.run();
      } catch (DesiredException e) {
         return;
      }
     fail("Desired exception not thrown. aaagh!");
   });
});
```

Each of these is plausible (though the last one sucks a bit), but not quite as neat as `@Test(expected=SomeException.class) public void myTest() {}` that you get in JUnit. Here are two motivations for some alternative:

- A terse syntax for expecting exceptions that's part of the spec
- Ability to define exception expectations hierarchically - i.e. group together in a `context` or `describe` a bunch of things that end in the same exception

There are a couple of ways I could imagine doing this and I'd like feedback on which might be desirable:

- As a `BlockConfiguration` as part of the `with` syntax
- Using a hook, a bit like `let` to provide the expected exception similar to `ExpectedException` but more _native_ to `Spectrum`

## Block Configuraton Approach

This is hierarchical and can be applied anywhere in the tree

e.g.

```java
describe("Suite", () -> {
   // all children are expected to thrown this exception, though they can specialise it
   // or even override it using the same `with` mechanism
   describe("throwing suite", with(expectedException(NullPointerException.class), () -> {
      it("Should throw", () -> {
          throw new NullPointerException();
      });
   }));

   it("doesn't have to throw", () -> {});

   it("requires its own exception", with(expectedException(RuntimeException.class, "boom!"), () -> {
       throw new RuntimeException("boom!");
   }));
});

```

## Hook approach

Let's use `ExpectedException` within Spectrum, or a clone of it if we don't want it to be _too_ JUnit. What we end up with will be similar to `let` but will add an expected exception hook instead of a `let hook`.

[...]

PR #123, opened 2017-09-07
POC for expected exception implementation.

ashleyfrieze:
A potential implementation for #121 

This was a POC I put together, provided as a PR for discussion.

I agree with the idea that things like this are done well with assertion frameworks. My experience with JUnit and AssertJ suggests that _it takes all sorts_. In that some things are better expressed in assertions and some things read easier another way.

So

```java
expect(object::method)
    .to(
        raise(SomeException.class));

// vs

exceptionExpectations.expect(SomeException.class);
object.method();
```

Both read well

### DIFF
--- a/docs/ExceptionHandling.md
+++ b/docs/ExceptionHandling.md
@@ -1,0 +1,60 @@
+# Exception handling in tests
+
+Often in unit testing, we want to check that the system under test throws an error in a certain situation.
+
+In native JUnit there are a couple of ways you can specify the expected exceptions of a test:
+
+```java
+// tag the test like this
+@Test(expected=SomeException.class)
+public void throwingTest() {}
+
+// or use the expected exception rule in the test class
+@Rule
+public ExpectedException expected = ExpectedException.none();
+
+@Test
+public void throwingTest() {
+  expected.expect(RuntimeException.class);
+}
+```
+
+These are great. **They do not work with Spectrum!!!**
+
+Even though Spectrum supports JUnit `@Rule` both in the test class and with the `junitMixin` function, the JUnit `ExpectedException` mechanism does not work. This is because the test has already failed owing to Spectrum's error handling before the JUnit exception rule gets a chance to say it's _ok_.
+
+## Use assertion libraries instead
+
+Your favourite assertion library may already solve this problem. Spectrum does not require or recommend an assertion library, so you can use whatever you with. AssertJ's `assertThatThrownBy` is a good fit for this.
+
+E.g.
+
+```java
+it("expects a boom", () -> {
+   assertThatThrownBy(() -> { throw new Exception("boom!"); }).isInstanceOf(Exception.class)
+                                                             .hasMessageContaining("boom");
+}
+```
+
+## Spectrum Expected Exceptions
+
+Spectrum provides the following syntax for expecting exceptions in your specs:
+
+```java
+describe("Some suite", () -> {
+  // in the parent of the test in the hierarchy, you add a call to `expectExceptions()`
+  // this gives you an `ExceptionExpectation` object you can set expectations in
+  ExceptionExpectation expectation = expectExceptions();
+  it("correctly goes bang", () -> {
+    // set the expectations here
+    expectation.expectMessage("bang");
+    throw new RuntimeException("bang");
+  });
+});
+```
+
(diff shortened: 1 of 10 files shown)


docs/ExceptionHandling.md
```diff
@@ -0,0 +1,60 @@
+# Exception handling in tests
```
ashleyfrieze: Whichever way we go, we probably ought to describe the limitations of what we have and the recommended route for exception expectation.

src/main/java/com/greghaskins/spectrum/ExceptionExpectation.java
```diff
@@ -0,0 +1,45 @@
+package com.greghaskins.spectrum;
+
+import java.util.function.Predicate;
+
+/**
+ * Allows expectations for exceptions to be specified by a spec.
+ */
+public interface ExceptionExpectation {
```
ashleyfrieze: I bet this looks similar to some of your work @greghaskins

src/main/java/com/greghaskins/spectrum/ExceptionExpectation.java
```diff
@@ -0,0 +1,45 @@
...
+   * @param typeOfException the thrown exception must be of this exact type.
+   */
+  void expect(final Class<?> typeOfException);
+
+  /**
+   * General purpose expectation. "and"ed to all other expectations called on the object.
+   * @param description description of the expectation for error reporting
+   * @param exceptionIsExpected defines a single check on the exception that must be valid
+   */
+  void expect(final String description, final Predicate<Throwable> exceptionIsExpected);
```
ashleyfrieze: I was quite proud of this idea - makes the whole thing way more extensible.

src/main/java/com/greghaskins/spectrum/internal/Suite.java
```diff
@@ -112,14 +113,29 @@ private void inheritConfigurationFromParent(final BlockConfiguration fromParent)
   }
 
   private Child configuredChild(final Child child, final Block block) {
-    merge(this.configuration.forChild(), ConfiguredBlock.configurationFromBlock(block))
+    merge(this.configuration.forChild(), configurationFromBlock(block))
```
ashleyfrieze: static import nicety

src/main/java/com/greghaskins/spectrum/internal/exception/ExceptionConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+      ((LeafChild) child).addLeafHook(expectationProxy.asHook(), HookContext.Precedence.ROOT);
+    }
+  }
+
+  @Override
+  public BlockConfigurable<ExceptionConfiguration> merge(BlockConfigurable<?> other) {
+    // always supersedes higher level definition
+
+    return this;
+  }
```
ashleyfrieze: This is similar to the `BlockTimeout` class, so there's a common base class to extract.

src/main/java/com/greghaskins/spectrum/internal/exception/ExceptionExpectationProxy.java
```diff
@@ -0,0 +1,77 @@
...
+import com.greghaskins.spectrum.internal.hooks.Hook;
+import com.greghaskins.spectrum.internal.hooks.NonReportingHook;
+
+import java.util.function.Predicate;
+import java.util.function.Supplier;
+
+/**
+ * Proxy to the exception expectation supplier.
+ */
+public class ExceptionExpectationProxy implements ExceptionExpectation {
```
ashleyfrieze: If we merge the _proxy code_ I may be able to dispense with this proxy class... I think I'd like to do that.

src/test/java/specs/ExpectedExceptionSpecs.java
```diff
@@ -0,0 +1,251 @@
...
+import org.junit.Rule;
+import org.junit.rules.ExpectedException;
+import org.junit.runner.Result;
+import org.junit.runner.RunWith;
+
+import java.io.IOException;
+import java.util.function.Supplier;
+
+@RunWith(Spectrum.class)
+public class ExpectedExceptionSpecs {
```
ashleyfrieze: I hope this class will explain the thinking...

ashleyfrieze:
Apparently there's a code coverage penalty from this implementation... consider that something I'll fix if we proceed with it. IntelliJ measured 100% on that code.. D'oh!

Issue #47, opened 2016-11-09
Support running tests in a random order?


tjarratt:
Hey friends!

I rather like this project a lot -- it makes it very easy to write BDD style tests in Java. Many thanks for writing and maintaining it.

I care very deeply about test ordering and randomization - rspec, jasmine, cedar, quick, etc all support test randomization as a way of detecting test pollution.

Is this a feature that is already in Spectrum? I couldn't find any documentation or code to that provided any hint as to how to do this, so I have to assume it doesn't work.

Would this be a desired feature, if I possibly found the time to contribute a PR? Any thoughts as to how it might work?

Thank you!

greghaskins:
I'm glad Spectrum has been helpful for you, @tjarratt . Thanks for the feedback, it helps motivate us to keep working on it. 

Spectrum doesn't currently support randomization, you're right about that. I'm definitely open to the idea, though. A PR would be great if you get the time and are interested in seeing this through. In general, our approach has been to follow the example of RSpec and Jasmine, so we'd want Spectrum to behave in a similar way. I often use their examples and even test cases to match behavior. 

It looks like both [RSpec](https://relishapp.com/rspec/rspec-core/v/3-5/docs/command-line/order) and [Jasmine](https://jasmine.github.io/2.5/node.html) use CLI flags and config files to activate randomization. How would you propose activating/configuring the random ordering (and seed)? Spectrum currently doesn't have any kind of project-wide config or CLI interface. I imagine you could do it file-by-file with an annotation parameter like `@SpectrumOptions(random=true)` or something, but it seems like something you'd want to activate randomization globally (on CI, for example). Perhaps Spectrum could use an environment variable or system property?

tjarratt:
Sorry to took me two weeks to get back to this idea, I'm still interested in getting a PR together.

As a very-infrequent contributor to Jasmine, it warms my heart to hear that other people in the BDD community look to it's test cases for matching behavior. That is literally awesome. Thank you.

I think you're absolutely right on a few points here

* most use cases need global randomization (especially on CI)
* it *could* be done file by file, but...
* using a system property would probably be simplest, since it would integrate well with `gradle`, `maven` (and presumably `sbt`, although I haven't used it yet).

So my proposal is to introduce two specific system properties:

`spectrum.randomize` -- this would be a boolean, defaulting to true
`spectrum.seed` -- an integer value, defaulting to the current number of seconds since unix epoch

[...]

PR #127, opened 2017-09-22
Consider patch release branch: Gherkin glitches

ashleyfrieze:
Fixes for inconsistencies in hook and configuration behaviour with Gherkin Syntax

PR #128, opened 2017-09-25
Make randomOrder configuration, allowing tests to be run in any order…

ashleyfrieze:
… within their part of the tree.

A solution to #47 

@tjarratt - what do you think?

### DIFF
--- a/src/test/java/specs/RandomOrderSpecs.java
+++ b/src/test/java/specs/RandomOrderSpecs.java
@@ -1,0 +1,120 @@
+package specs;
+
+import static com.greghaskins.spectrum.Configure.randomOrder;
+import static com.greghaskins.spectrum.Configure.with;
+import static com.greghaskins.spectrum.Spectrum.describe;
+import static com.greghaskins.spectrum.Spectrum.it;
+import static com.greghaskins.spectrum.dsl.gherkin.Gherkin.*;
+import static java.util.stream.Collectors.toList;
+import static junit.framework.TestCase.fail;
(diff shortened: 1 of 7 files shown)


marioluan:
I'm really looking forward to having this feature pushed.

@greghaskins @tjarratt Any changes for this to get merged?

ashleyfrieze:
I would like to retest this with the gherkin style tests to make sure it can’t scramble them.

src/test/java/specs/RandomOrderSpecs.java
```diff
@@ -0,0 +1,120 @@
...
+          });
+          and("third step adds third", () -> {
+            strings.add("third");
+          });
+          then("the order is correct", () -> {
+            assertThat(strings, contains("first", "second", "third"));
+          });
+        });
+      }));
+    });
```
greghaskins: @ashleyfrieze Do you imagine needing more than these specs here ☝️ to sufficiently test the Gherkin-style DSL? This scenario seems pretty straightforward to me. I guess you might have a mixed `describe` where some pieces run in random order, but the right things (given/when/then) stay in the right order.

ashleyfrieze: These do look like they cover the scenario. I've a funny feeling there was something else worrying me about the feature, perhaps relating back to the bug I found in the gherkin tests. If you're going to merge PR #127, why don't we review this again after rebasing?

greghaskins: Sounds good. We'll look at it again after #127

PR #133, opened 2018-03-08
Fix Issue #132 -- Make `let` Lazily Evaluated

### LINKED ISSUE #132: Bug: `let()` is invoked before `beforeEach`
In RSpec, `let` blocks are evaluated and cached lazily in each spec only when their value is first needed within a given test. But, in Spectrum, it appears that the `let` blocks all fire off and get cached at the start of the test. In fact, they all actually seem to get invoked _before_ any `beforeEach` blocks. This means that `let` blocks that depend on any state that the `beforeEach` blocks manipulate will not be correct.

Here is a test that demonstrates the issue:
```Java
package com.rosieapp.test;

import static com.greghaskins.spectrum.dsl.specification.Specification.beforeEach;
import static com.greghaskins.spectrum.dsl.specification.Specification.context;
import static com.greghaskins.spectrum.dsl.specification.Specification.it;
import static com.greghaskins.spectrum.dsl.specification.Specification.let;
import static org.assertj.core.api.Assertions.assertThat;

import com.greghaskins.spectrum.Spectrum;
import com.greghaskins.spectrum.Variable;
import java.util.function.Supplier;
import org.junit.runner.RunWith;

@RunWith(Spectrum.class)
public class BadTest {
  {
    final Variable<String> theVariable = new Variable<>();
    final Supplier<String> result      = let(() -> theVariable.get());

    context("when `theVariable` is `x`", () -> {
      beforeEach(() -> {
        theVariable.set("x");
      });

      it("stores `x` in `result`", () -> {
        // Fails: `result` is `null` because it was fetched & cached before the `beforeEach`
        assertThat(result.get()).isEqualTo("x");
      });
    });

    context("when `theVariable` is `y`", () -> {
      beforeEach(() -> {
        theVariable.set("y");
      });

      it("stores `y` in `result`", () -> {
        // Fails: `result` is `x` because it was fetched & cached before the `beforeEach`
        assertThat(result.get()).isEqualTo("y");
      });
    });
  }
}
```

To work around this limitation, I'm having to use `beforeAll()` at the start of each context. That works (and is a bit more efficient since we only have to reset the variable once per context), but introduces the possibility that each of the tests within a context are sharing more state than desired.

GuyPaddock:
The previous implementation would cause all `let` blocks to be evaluated at the start of the test rather than on an as-needed basis, which is inconsistent with the behavior developers who are [used to RSpec](https://rspec.info/documentation/3.5/rspec-core/RSpec/Core/MemoizedHelpers/ClassMethods.html#let-instance_method) would expect. Now, `let` only evaluates its block when the test first needs a value, allowing the `let` block to depend on state that is being setup by other blocks (`beforeEach`, other `let` blocks, etc).

Despite this, there was value in preserving the old implementation as Spectrum's equivalent to [`let!` from RSpec](https://rspec.info/documentation/3.5/rspec-core/RSpec/Core/MemoizedHelpers/ClassMethods.html#let!-instance_method). As such, I've moved that implementation over to a new DSL helper called `eagerLet`. One use case where this is useful is defining values that can be used in a `beforeEach` block.

Closes #132.

### DIFF
--- a/docs/VariablesAndValues.md
+++ b/docs/VariablesAndValues.md
@@ -12,10 +12,14 @@ when the test is broken into separate steps.
 The `let` function is used to initialise a fresh, isolated, object for each spec.
 
 ### Common Variable Initialization
+#### Let
+The `let` helper function makes it easy to initialize common variables that are used in multiple
+specs. In standard JUnit you might expect to use the initializer list of the class or a `@Before`
+method to achieve the same. As there is no easy way for `beforeAll` or `beforeEach` to instantiate
+a value that will be used in the specs, `let` is the tool of choice.
 
-The `let` helper function makes it easy to initialize common variables that are used in multiple specs. In standard JUnit you might expect to use the initializer list of the class or a `@Before`	method to achieve the same. As there is no easy way for `beforeAll` or `beforeEach` to instantiate a value that will be used in the specs, `let` is the tool of choice.
-
-Values are cached within a spec, and lazily re-initialized between specs as in [RSpec #let](http://rspec.info/documentation/3.5/rspec-core/RSpec/Core/MemoizedHelpers/ClassMethods.html#let-instance_method).
+Values are cached within a spec, and lazily re-initialized between specs as in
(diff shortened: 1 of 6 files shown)


greghaskins:
Thanks for the contribution @GuyPaddock. This looks great,

